PR #6, opened 2016-08-05
CL21: added CL_DEVICE_HALF_FP_CONFIG define

bsochack:


jrprice:
For this one I just pushed `master` to `opencl21` to avoid them unnecessarily diverging. For some reason the hash for the commit in this branch was different to that in master.

PR #20, opened 2017-05-03
provide patch for PowerPC where 'vector' is unset

cdeterman:
While developing on a PowerPC cluster I discovered an error when including the cl_platform.h file complaining about 'vector' not naming a type.  I discovered in this [issue](https://github.com/bmwcarit/meta-ros/issues/355) referencing opencv having the same problem.  The following patch fixes this problem.

### DIFF
--- a/cl_platform.h
+++ b/cl_platform.h
@@ -408,14 +408,25 @@ typedef unsigned int cl_GLenum;
 
 /* Define basic vector types */
 #if defined( __VEC__ )
-   #include <altivec.h>   /* may be omitted depending on compiler. AltiVec spec provides no way to detect whether the header is required. */
-   typedef vector unsigned char     __cl_uchar16;
-   typedef vector signed char       __cl_char16;
-   typedef vector unsigned short    __cl_ushort8;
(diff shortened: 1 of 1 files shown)


jrprice:
Can you explain why this needs to be conditional on `defined(__powerpc__)`? The [patch that Debian uses](https://anonscm.debian.org/cgit/pkg-opencl/khronos-opencl-headers.git/tree/debian/patches/use__vector.patch) is to rename `vector` to `__vector` for *all* targets, including PowerPC.

Also, I'm afraid we've restructured the repository since you submitted this patch, so it doesn't merge cleanly. All OpenCL versions are now present in a single branch, which means changes like this should be made for every version in a single commit (ideally).

sergiomb2:
yes , IMHO the Debian patch is the correct patch also used here:  https://github.com/openembedded/meta-openembedded/commit/97dfbbbf6bddbd360638cde98cb53790ccc83059

jrprice:
Merged the approach used by the Debian patch in 1da98fd.

Issue #17, opened 2017-03-10
inaccurate mathematical constants


ian-bruce:
<cl_platform.h> defines some standard mathematical constants, in both single- and double-precision formats:

    $ ( sed -n '/stdint/,$p' | grep 'CL_M_' ) <cl_platform.h

    #define  CL_M_E             2.718281828459045090796
    #define  CL_M_LOG2E         1.442695040888963387005
    #define  CL_M_LOG10E        0.434294481903251816668
    #define  CL_M_LN2           0.693147180559945286227
    #define  CL_M_LN10          2.302585092994045901094
    #define  CL_M_PI            3.141592653589793115998
    #define  CL_M_PI_2          1.570796326794896557999
    #define  CL_M_PI_4          0.785398163397448278999
    #define  CL_M_1_PI          0.318309886183790691216
    #define  CL_M_2_PI          0.636619772367581382433
    #define  CL_M_2_SQRTPI      1.128379167095512558561
    #define  CL_M_SQRT2         1.414213562373095145475
    #define  CL_M_SQRT1_2       0.707106781186547572737
    #define  CL_M_E_F           2.71828174591064f
    #define  CL_M_LOG2E_F       1.44269502162933f
    #define  CL_M_LOG10E_F      0.43429449200630f
    #define  CL_M_LN2_F         0.69314718246460f
    #define  CL_M_LN10_F        2.30258512496948f
    #define  CL_M_PI_F          3.14159274101257f
    #define  CL_M_PI_2_F        1.57079637050629f
    #define  CL_M_PI_4_F        0.78539818525314f
    #define  CL_M_1_PI_F        0.31830987334251f
    #define  CL_M_2_PI_F        0.63661974668503f
    #define  CL_M_2_SQRTPI_F    1.12837922573090f
    #define  CL_M_SQRT2_F       1.41421353816986f
    #define  CL_M_SQRT1_2_F     0.70710676908493f

Bizarrely, the single- and double-precision versions **DO NOT MATCH BEYOND THE SEVENTH DIGIT**:

    $ ( sed -n '/stdint/,$p' | grep 'CL_M_' ) <cl_platform.h |
        ( sort | sed -r -e 's/ 0\./0./' -e 's/(.{36})(.*)/\1 \2/' )

    #define  CL_M_1_PI         0.3183098 86183790691216
    #define  CL_M_1_PI_F       0.3183098 7334251f

    #define  CL_M_2_PI         0.6366197 72367581382433
[...]

PR #22, opened 2017-07-18
Use correctly rounded decimal mathematical constants

jrprice:
This is the proposed change discussed in GitHub issue #17, as well as internally inside the working group. We don't yet have unanimous consensus to make this change, but having a PR to look at will hopefully help.


Dithermaster:
The calculator I have handy only goes to float precision, but these all look good to me. I verified they only changed in the digits that varied before from published constants. Finally, I used the before and after numbers in a Visual Studio 2013 C++ program and confirmed that the float and double values created were equivalent. LGTM.

jrprice:
Working group agreed on July 25th call to accept this.

PR #24, opened 2017-12-29
Add cl_khr_il_program to OpenCL 1.2 and 2.0 headers

pierremoreau:
The “cl_khr_il_program” extension is present in both the OpenCL 1.2 and 2.0 Extensions Specification, but could not be found in the “cl_ext.h” file.

### DIFF
--- a/opencl12/CL/cl_ext.h
+++ b/opencl12/CL/cl_ext.h
@@ -519,6 +519,40 @@ clSetKernelExecInfoARM(cl_kernel            /* kernel */,
 
 #endif /* CL_VERSION_1_2 */
 
+
+/*******************************
+ * cl_khr_il_program extension *
+ *******************************/
+
+#ifndef cl_khr_il_program
+#define cl_khr_il_program 1
+
+/* New property to clGetDeviceInfo for retrieving supported intermediate
+ * languages
+ */
+#define CL_DEVICE_IL_VERSION_KHR                    0x105B
+
+/* New property to clGetProgramInfo for retrieving for retrieving the IL of a
+ * program
+ */
+#define CL_PROGRAM_IL_KHR                           0x1169
+
+extern CL_API_ENTRY cl_program
+  CL_API_CALL clCreateProgramWithILKHR(
+      cl_context /* context */,
+      const void * /* il */,
+      size_t /* length */,
+      cl_int * /* errcode_ret */);
+
+typedef CL_API_ENTRY cl_program
+  (CL_API_CALL *clCreateProgramWithILKHR_fn)(
+      cl_context /* context */,
+      const void * /* il */,
+      size_t /* length */,
+      cl_int * /* errcode_ret */) CL_EXT_SUFFIX__VERSION_1_2;
+
+#endif /* cl_khr_il_program */
+
 #ifdef __cplusplus
 }
 #endif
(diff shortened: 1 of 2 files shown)


opencl12/CL/cl_ext.h
```diff
@@ -519,6 +519,40 @@ clSetKernelExecInfoARM(cl_kernel            /* kernel */,
...
+ * program
+ */
+#define CL_PROGRAM_IL_KHR                           0x1169
+
+extern CL_API_ENTRY cl_program
+  CL_API_CALL clCreateProgramWithILKHR(
+      cl_context /* context */,
+      const void * /* il */,
+      size_t /* length */,
+      cl_int * /* errcode_ret */);
```
gwawiork: I am wondering if CL_EXT_SUFFIX__VERSION_1_2 should be placed here like in other definitions.

opencl20/CL/cl_ext.h
```diff
@@ -553,6 +553,40 @@ clImportMemoryARM( cl_context context,
...
+ * program
+ */
+#define CL_PROGRAM_IL_KHR                           0x1169
+
+extern CL_API_ENTRY cl_program
+  CL_API_CALL clCreateProgramWithILKHR(
+      cl_context /* context */,
+      const void * /* il */,
+      size_t /* length */,
+      cl_int * /* errcode_ret */);
```
gwawiork: I am wondering if CL_EXT_SUFFIX__VERSION_1_2 should be placed here like in other definitions.

pierremoreau: I was wondering, should it be CL_EXT_SUFFIX__VERSION_1_2 or CL_EXT_SUFFIX__VERSION_2_0?

gwawiork: CL_EXT_SUFFIX__VERSION_1_2 here as well

gwawiork:
I do not have write access to the repo, please merge and squash it yourself or ask somebody else.

pierremoreau:
I do not have merge rights myself either, so I’ll squash them and wait for someone with merge rights to come and have a look.
Thank you for having a look @gwawiork!

jrprice:
Sorry for the delay - squashed and merged in b5c3680.

Thanks for the contribution!

pierremoreau:
No worries @jrprice , and thanks for merging.
By the way, I was wondering how it is supposed to work in OpenCL 2.1 and 2.2. From what I have read in the extension specification, even if an extension makes it to core, it should still be accepted by implementations. In that case, do we need to update cl_ext.h for OpenCL 2.1 and 2.2, to add that extension there as well?

jrprice:
> From what I have read in the extension specification, even if an extension makes it to core, it should still be accepted by implementations. In that case, do we need to update cl_ext.h for OpenCL 2.1 and 2.2, to add that extension there as well?

Good question. I *think* we should indeed add this to the 2.1/2.2 headers as you suggest, but I'm not 100% sure. @bashbaug do you know what our stance is for extensions that move to core like this?

bashbaug:
> do you know what our stance is for extensions that move to core like this?

Good question.  Personally I tend to agree that the extension headers should be a union of all OpenCL extensions, but we should take this question to the OpenCL working group to see if there's a reason why some extensions are missing from some OpenCL header versions.

Issue #34, opened 2018-10-31
warning C4201: nonstandard extension used: nameless struct/union


MathiasMagnus:
I guess this is the same issue as was the case with GCC inside `typedef struct _cl_image_desc` for eg. which was circumvented with the code:
```c++
#ifdef __GNUC__
    __extension__   /* Prevents warnings about anonymous union in -pedantic builds */
#endif
    union {
      cl_mem                  buffer;
      cl_mem                  mem_object;
    };
```
MSVC has exactly the same problem when compiled with [/Za, strict ANSI conformance](https://docs.microsoft.com/hu-hu/cpp/error-messages/compiler-warnings/compiler-warning-level-4-c4201?view=vs-2017).

MathiasMagnus:
By the way, would giving this enum a name break the ABI? Or are there other implications? Isn't it more elegant to actually conform to ANSI C than to disable extensions?

jrprice:
It wouldn't break the ABI, but it would potentially break existing code, which is undesirable. We'll certainly consider avoiding anonymous unions like this in the future though.

I've merged PR #35, so closing this issue now.

PR #35, opened 2018-10-31
Fixed w4201 triggering with MSVC in /W4 /Za builds

MathiasMagnus:
This PR fixes issue #34


jrprice:
Thanks for the fix - rebased locally and pushed to master.

Issue #33, opened 2018-10-15
Missing _KHR suffix from mipmap extension tokens


oddhack:
In https://github.com/KhronosGroup/OpenCL-Registry/pull/55 I have a reference page fix to match the 2.2 extension spec, which adds a missing _KHR suffix to several tokens introduced by the mipmap extension. The suffix also appears to be missing from (all of) the headers in this repo. If the spec is taken as the controlling document, we should probably add aliases of the non-suffixed tokens with the suffix added, and maybe acknowledge somewhere (ref pages? extension spec?) that the non-suffixed versions were an error.

bashbaug:
I haven't done the full spec archaeology, but I suspect mipmaps were initially a core feature before being moved to an extension.  During this process the enums were never updated with a KHR suffix.  Note that the three enums in question (`CL_SAMPLER_MIP_FILTER_MODE`, `CL_SAMPLER_LOD_MIN`, and `CL_SAMPLER_LOD_MAX`) are in the main cl.h header file, protected by a `CL_VERSION_2_0` #ifdef, and are not in cl_ext.h where one would expect to find enums for a mipmap extension.

To resolve this, I propose:

1. Add suffixed variants of the three enums to cl_ext.h.
1. Add a comment to the un-suffixed enums in cl.h acknowledging that the un-suffixed enums are an error but are maintained for backwards compatibility.

If this sounds OK I'll create a PR with these changes.

Note: There are two un-suffixed enums in cl_ext.h for cl_khr_spir: `CL_DEVICE_SPIR_VERSIONS` and `CL_PROGRAM_BINARY_TYPE_INTERMEDIATE`.  Should we add suffixed aliases for these enumerants, also?

oddhack:
If the WG likes that, I can just go ahead and put in my corresponding changes to the refpages, so it would be good for me. I think any other suffix errors like this should be treated similarly (and there would need to be corresponding fixes to the refpages, too).

jrprice:
Addressed in 03490c3.

PR #36, opened 2018-10-31
add suffixed enums for cl_khr_mipmap_image

bashbaug:
This change adds KHR suffixed enums to cl_ext.h for the cl_khr_mipmap_image extension, and a comment describing why the un-suffixed enums in the main cl.h are incorrect (but are left for compatibility).

See:
https://github.com/KhronosGroup/OpenCL-Headers/issues/33#issuecomment-429996700


oddhack:
I went ahead and merged the corresponding change to the ref pages.

Issue #42, opened 2019-01-17
enums for cl_khr_image2d_from_buffer


bashbaug:
While making the changes in #41 I noticed that we do not currently define enums for `cl_khr_image2d_from_buffer` in `cl_ext.h`.  Instead, the enums for this extension are in `cl.h`, protected by `#ifdef CL_VERSION_1_2`, and do not have the `KHR` suffix.  This doesn't seem correct.

One solution is to add suffixed versions of the enums to `cl_ext.h`, similar to what was done in #33 for the mipmap extension.  Note however, the extensions doc currently does not include a `KHR` suffix on the enums, so this may not be the right thing to do in this case.

Alternatively, we could add the un-suffixed enums to `cl_ext.h` if they aren't defined in `cl.h`, similar to what was done for subgroups (see #2).  This solution preserves source backwards compatibility.

bashbaug:
In the March 5th teleconference we decided to:

* Add `KHR` suffixed versions of these enums and add to `cl_ext.h`.
* Change the `#ifdef` in `cl.h` so the un-suffixed enums are protected by `CL_VERSION_2_0`, not `CL_VERSION_1_2`, since the extension became a core feature in OpenCL 2.0.
* Update the extension spec to refer to the `KHR` suffixed enums and not the un-suffixed enums.

I will create a PR in this repo for the first two changes and a PR in the spec repo for the last change.

bashbaug:
Both the spec change and the headers change are merged, so I am closing this issue.

PR #44, opened 2019-03-12
added suffixed enums for cl_khr_image2d_from_buffer

bashbaug:
This is a partial fix for #42.

This change adds `KHR` suffixed enums for `cl_khr_imaged2d_from_buffer` to cl_ext.h.

It also moves the un-suffixed enums for `cl_khr_image2d_from_buffer` in cl.h to a section protected by `#ifdef CL_VERSION_2_0`, not `#ifdef CL_VERSION_1_2`, since the extension became a core feature in OpenCL 2.0.

### DIFF
--- a/CL/cl.h
+++ b/CL/cl.h
@@ -356,10 +356,10 @@ typedef struct _cl_buffer_region {
 #define CL_DEVICE_REFERENCE_COUNT                        0x1047
 #define CL_DEVICE_PREFERRED_INTEROP_USER_SYNC            0x1048
 #define CL_DEVICE_PRINTF_BUFFER_SIZE                     0x1049
-#define CL_DEVICE_IMAGE_PITCH_ALIGNMENT                  0x104A
-#define CL_DEVICE_IMAGE_BASE_ADDRESS_ALIGNMENT           0x104B
 #endif
 #ifdef CL_VERSION_2_0
+#define CL_DEVICE_IMAGE_PITCH_ALIGNMENT                  0x104A
+#define CL_DEVICE_IMAGE_BASE_ADDRESS_ALIGNMENT           0x104B
 #define CL_DEVICE_MAX_READ_WRITE_IMAGE_ARGS              0x104C
 #define CL_DEVICE_MAX_GLOBAL_VARIABLE_SIZE               0x104D
 #define CL_DEVICE_QUEUE_ON_DEVICE_PROPERTIES             0x104E
--- a/CL/cl_ext.h
+++ b/CL/cl_ext.h
@@ -147,20 +147,28 @@ typedef CL_API_ENTRY cl_program
                                            size_t       length,
                                            cl_int *     errcode_ret) CL_EXT_SUFFIX__VERSION_1_2;
 
-/* Extension: cl_khr_image2D_buffer
+/* Extension: cl_khr_image2d_from_buffer
  *
- * This extension allows a 2D image to be created from a cl_mem buffer without a copy.
- * The type associated with a 2D image created from a buffer in an OpenCL program is image2d_t.
- * Both the sampler and sampler-less read_image built-in functions are supported for 2D images
- * and 2D images created from a buffer.  Similarly, the write_image built-ins are also supported
- * for 2D images created from a buffer.
+ * This extension allows a 2D image to be created from a cl_mem buffer without
+ * a copy. The type associated with a 2D image created from a buffer in an
+ * OpenCL program is image2d_t. Both the sampler and sampler-less read_image
+ * built-in functions are supported for 2D images and 2D images created from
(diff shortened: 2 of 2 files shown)


Smeany:
There are CPUs (Intel [https://compubench.com/device.jsp?benchmark=compu15d&os=OS%20X&api=cl&D=Intel%28R%29+Core%28TM%29+i3-3250+CPU&testgroup=info]) that have support for OpenCL 1.2 and the mentioned extension. I don't see why the extention becoming a feature in 2.0 justifies exclusion of the headers from prior versions.

Smeany:
It gets even worse when trying to implement zero copy image from buffer with the c++ headers in OpenCL 1.2 as it requires you to set the _cl_image_desc_ structs _buffer_/_mem_object_ union to the desired buffer object that needs to be transformed into an image [(like in this example)](https://www.intel.com/content/dam/develop/external/us/en/documents/imagefrombuffer-531122.pdf). The C++ Image2d Implementation simply uses the c versions clCreateImage function but with a fixed _cl_image_desc_ which in turn doesn't let you specify the _buffer_/_mem_object_ union...

bashbaug:
Hi @Smeany, regarding:

> I don't see why the extention becoming a feature in 2.0 justifies exclusion of the defines from versions prior to 2.0.

The basic idea is that you would use the `KHR`-suffixed versions in `cl_ext.h` for devices that supports the extension and the non-suffixed versions in `cl.h` for devices that supports the core feature.

This is slightly academic in that all devices that support the core feature also must support the extension, and both the suffixed and un-suffixed enums have the same value.  For maximal compatibility you could use the suffixed enums exclusively - this would be my recommendation unless you're compiling for OpenCL 2.0 or newer devices.

I'm not quite following the issue with `cl_image_desc` and the C++ bindings, but please feel free to file an issue in the [OpenCL-LHPP](https://github.com/KhronosGroup/OpenCL-CLHPP) repo if you have ideas for improvement.  This will likely get more attention than here, since this PR is merged.

Thanks!

Smeany:
Thank you @bashbaug , I was not aware of the `cl_ext.h` header.
The problem I tried to solve revolves around finding a way to enable read and write access to an image2d object, but opencl 1.2 limits the acces to either read or write and doesn't support both at the same time. So the workaround I came across was to just use a buffer object instead for r/w capability, which can then be turned into an image2d object via the `cl_khr_image2d_from_buffer ` extension without the need for an unnecessary copy.
Following the [description](https://www.intel.com/content/dam/develop/external/us/en/documents/imagefrombuffer-531122.pdf) in step 3 before creating the image from a buffer you have to assign the `cl_buffer` object to a `cl_image_desc` object and its `mem_object` property. This `cl_image_desc` object is then used to call the C versions `clCreateImage`. As I've written everything in C++ and wanted to continue doing so I looked for a way to access the `cl_image_desc` property of the C++ Image2d until I stumbled upon the Sourcecode for the Image2d C++ Constructor for opencl 1.2.
It creates it's own intermediary `cl_image_desc` object and theres no overloaded Image2d Constructor which would allow assigning a custom `cl_image_desc` object. To come around this I had to use the C versions  `clCreateImage` function and then wrap it into a C++ Object. Correct me if I'm wrong, but it seems that theres no other way around using C.
```
#if CL_HPP_TARGET_OPENCL_VERSION >= 120
        if (useCreateImage)
        {
            cl_image_desc desc = {0};
            desc.image_type = CL_MEM_OBJECT_IMAGE2D;
            desc.image_width = width;
            desc.image_height = height;
            desc.image_row_pitch = row_pitch;

            object_ = ::clCreateImage(
                context(),
                flags,
                &format,
                &desc,
                host_ptr,
                &error);

            detail::errHandler(error, __CREATE_IMAGE_ERR);
            if (err != NULL) {
                *err = error;
            }
        }
#endif // CL_HPP_TARGET_OPENCL_VERSION >= 120```

**Comment by bashbaug** [comment]
> it seems that theres no other way around using C.

Could you use this Image2D constructor?

https://github.com/KhronosGroup/OpenCL-CLHPP/blob/master/include/CL/opencl.hpp#L4856

```c++
    /*! \brief Constructs a 2D Image from a buffer.
    * \note This will share storage with the underlying buffer.
    *
    *  Wraps clCreateImage().
    */
    Image2D(
        const Context& context,
        ImageFormat format,
        const Buffer &sourceBuffer,
        size_type width,
        size_type height,
        size_type row_pitch = 0,
        cl_int* err = nullptr)
...
```

Note, you will need to define `CL_HPP_USE_CL_IMAGE2D_FROM_BUFFER_KHR` or compile for OpenCL 2.0 or newer to use this constructor, at least until https://github.com/KhronosGroup/OpenCL-CLHPP/pull/148 is merged.

PR #45, opened 2019-04-11
Simplify the definition of deprecation prefixes/suffixes

kpet:


kpet:
@jrprice Thanks for the review.

PR #47, opened 2019-05-09
Add CMake Packaging + CI For Package

mmha:
# Rationale

## State of using OpenCL in CMake
The canonical way of using OpenCL inside of a CMake-based project goes like this:
```cmake
add_executable(app main.cpp)
find_package(OpenCL REQUIRED)
target_link_libraries(app PRIVATE OpenCL::OpenCL)
```

This works because [Kitware added a `FindOpenCL.cmake`](https://gitlab.kitware.com/cmake/cmake/blob/master/Modules/FindOpenCL.cmake) to the standard CMake distribution which is executed during the `find_package` call. This comes with a few problems:
* `FindOpenCL.cmake` has to consider many different OpenCL implementations, that may or may not use an ICD. Therefore, it could only guess what the dependencies of the found `libOpenCL` are. As of now, there are no dependencies added at all.
* The Khronos ICD loader depends on `libdl` and `libpthread` on Linux. Just linking with `OpenCL::OpenCL` will likely result in undefined references.
* Looking at the big picture, we're depending on a third party (Kitware) to do the packaging for us. This shouldn't be the case. We, as vendors, know best what our code needs in order to be consumed by library users. Cut the middleman!

## What does this have to do with the Headers?
The obvious solution would be for the ICD loader to install an `OpenCLConfig.cmake` and put our dependencies there. However, since the headers are a separate project, they are an external dependency that must be referenced (or rather should; since we don't want to install the headers upon ICD installation). This can happen by shipping a `FindOpenCLHeader.cmake` along with the ICD or by adding CMake packaging to the Headers as well. The latter solution is what this PR is about.

## Other benefits
* Projects that just need the OpenCL headers, but not the ICD (such as OpenCL implementations) can now start to `find_package(OpenCLHeaders)` and keep the headers out of their tree.
* We now have CI in place that can be extended further. Same goes for the CMakeLists and the possibility to add proper tests.

## Implementation details
* The CMakeLists does not compile any code. It merely adds installation rules.
* CI is implemented by creating a Conan test package. The test package is built for different configurations. Since this uses Docker images, new compilers can easily be added  later on if desired.
* The README was improved to document how the headers are assumed to be consued.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,6 +2,7 @@ cmake_minimum_required(VERSION 3.0)
 project(OpenCLHeaders VERSION 2.2 LANGUAGES NONE)
 
 add_library(OpenCLHeaders INTERFACE)
+add_library(OpenCL::OpenCLHeaders ALIAS OpenCLHeaders)
 target_include_directories(OpenCLHeaders INTERFACE
 	$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>
 	$<INSTALL_INTERFACE:include>
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,42 @@
+cmake_minimum_required(VERSION 3.0)
+project(OpenCLHeaders VERSION 2.2 LANGUAGES NONE)
+
+add_library(OpenCLHeaders INTERFACE)
+target_include_directories(OpenCLHeaders INTERFACE
+	$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>
+	$<INSTALL_INTERFACE:include>
+)
+
+install(TARGETS OpenCLHeaders EXPORT OpenCLHeadersTargets)
+install(DIRECTORY "CL" DESTINATION include)
+
+export(EXPORT OpenCLHeadersTargets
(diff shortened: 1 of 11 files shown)


CMakeLists.txt
```diff
@@ -0,0 +1,43 @@
...
+
+export(EXPORT OpenCLHeadersTargets
+  FILE "${PROJECT_BINARY_DIR}/OpenCLHeaders/OpenCLHeadersTargets.cmake"
+  NAMESPACE OpenCL::
+)
+file(WRITE "${PROJECT_BINARY_DIR}/OpenCLHeaders/OpenCLHeadersConfig.cmake"
+	"include(\"\${CMAKE_CURRENT_LIST_DIR}/OpenCLHeadersTargets.cmake\")"
+)
+
+set(config_package_location lib/cmake/OpenCLHeaders)
```
sl1pkn07: better if is set in ${CMAKE_INSTALL_DATADIR} because is header only package

bashbaug:
@MathiasMagnus can we close this PR now that #170 has been merged?

MathiasMagnus:
I believe this one can be closed as well. #170 knows exactly this. Should the author request something more, the issue can be reopened.

PR #50, opened 2019-07-02
Add API function pointer and ICD dispatch table definitions

kpet:
...so that implementations (and others) can use definitions
straight from the headers instead of mirroring them.

The definitions come from https://github.com/KhronosGroup/OpenCL-ICD-Loader
with the following name changes:

- function pointers are now prefixed with 'cl_api_'
- the dispatch table type is now typedef'ed as 'cl_icd_dispatch', retaining
  compatibility with `struct _cl_icd_dispatch`.

Contributes to https://github.com/KhronosGroup/OpenCL-ICD-Loader/issues/72

### DIFF
--- a/CL/cl_icd.h
+++ b/CL/cl_icd.h
@@ -1,0 +1,1072 @@
+/*******************************************************************************
+ * Copyright (c) 2019 The Khronos Group Inc.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a
+ * copy of this software and/or associated documentation files (the
+ * "Materials"), to deal in the Materials without restriction, including
+ * without limitation the rights to use, copy, modify, merge, publish,
+ * distribute, sublicense, and/or sell copies of the Materials, and to
+ * permit persons to whom the Materials are furnished to do so, subject to
+ * the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included
+ * in all copies or substantial portions of the Materials.
+ *
+ * MODIFICATIONS TO THIS FILE MAY MEAN IT NO LONGER ACCURATELY REFLECTS
+ * KHRONOS STANDARDS. THE UNMODIFIED, NORMATIVE VERSIONS OF KHRONOS
+ * SPECIFICATIONS AND HEADER INFORMATION ARE LOCATED AT
+ *    https://www.khronos.org/registry/
+ *
+ * THE MATERIALS ARE PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+ * EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+ * MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+ * IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY
+ * CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT,
+ * TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+ * MATERIALS OR THE USE OR OTHER DEALINGS IN THE MATERIALS.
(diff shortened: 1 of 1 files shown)


kpet:
Thanks for the feedback!

>automatically generating code like the function pointers from the cl.xml

Wholeheartedly agree. Trying to maintain these definitions without a single source of truth is bound to result in mistakes.

>Is it possible that the function pointers could be useful by themselves, without the dispatch table? If so, consider putting them in separate files.

Yes, I can see them being useful on their own but my initial reaction would be that a separate file feels a bit overkill. Having to pull the dispatch table declaration as well doesn't feel like too much pollution. We may want to consider renaming the file from `cl_icd.h` to something more generic though. Any suggestions?

>I'd consider a different prefix ("_cl", perhaps?) for declarations that are not part of the official API, meaning things like the function pointers and the dispatch table.

I'm not so sure, the C rules for reserved identifiers would be lurking uncomfortably near. `_cl_pfn` would be valid but `_cl_PFN` or `__cl_pfn` would be reserved.

Regarding the dispatch table, I've created https://github.com/KhronosGroup/OpenCL-Docs/pull/101.

>Should this file #include cl.h, since it has a dependency on the types it defines?

Yes, it definitely should. Thanks for catching that.

>Consider adding a comment in this file and/or a section in the README to describe what this file contains.

Yes, this can't hurt. I've added a short summary at the top of the header file.

Kerilk:
Just a quick comment about the fact that some (usually open source POCL, SOCL, etc...) implementation of OpenCL use the old ```struct _cl_icd_dispatch``` structure type, which was provided in the ocl_icd.h file from the open source development packages.
If we want the new functionalities to serve as a drop-in replacement it may be useful to keep the naming unchanged.

Brice

bashbaug:
> > I'd consider a different prefix ("_cl", perhaps?) for declarations that are not part of the official API, meaning things like the function pointers and the dispatch table.
> 
> I'm not so sure, the C rules for reserved identifiers would be lurking uncomfortably near. `_cl_pfn` would be valid but `_CL_PFN` or `__cl_pfn` would be reserved.

Ah, good point.  Maybe `cl_api` then, to match the `CL_API` prefix already used for things like `CL_API_ENTRY` and `CL_API_CALL`?

I'm fine with the existing `cl_pfn`, or `_cl_icd` for compatibility with ocl-icd as well, so no action required, just enumerating options.

What are next steps for this PR?

kpet:
@Kerilk Noted. I'm guessing that `ocl_icd.h` file is a copy of the Khronos loader definition, isn't it? I don't think being a drop-in replacement is really a requirement and I think I'd rather these projects make a conscious decision to transition. That being said, we should probably change `ocl_icd.h` as well so that it uses the new definition from the headers. Where is this file maintained?

@bashbaug  Yeah, I prefer `cl_api` as well. I'll update the PR.

I think the next step is to merge if we're in agreement, isn't it?

Kerilk:
@kpet No, the ocl_icd.h is generated from a YAML file containing parsed entry points. The position of the entry points in the table was reversed engineered from vendor libOpenCL.so using dummy OpenCL implementations. This was before the sources were made public.

You can obtain it building the loader:
https://github.com/OCL-dev/ocl-icd

kpet:
I'd like to progress on this. Is the only thing keeping this from progressing the name change from `struct _cl_icd_dispatch`? If that's the only problem I'm happy to add an alias or even rename the structure. (cc @Kerilk @bashbaug)

Kerilk:
Fine by me.

bashbaug:
No issues from me either.

I have a slight preference for `typedef struct _cl_icd_dispatch { ... } cl_icd_dispatch;`, or even just `struct _cl_icd_dispatch { ... };`, since this is what is in the cl_khr_icd extension specification and what other open source projects are using, but I'm really fine either way.

Let me know when this is ready to go and I can merge.

kpet:
@bashbaug Done, PTAL.

bashbaug:
Looks good to me - thanks!

I still think we might want to separate out the function pointers from the dispatch table definition, especially if we start to generate the function pointers, but I think we could do that as a follow-on change if desired.

Issue #57, opened 2019-12-02
Provide conversion functions between cl_half and other types


kpet:
We currently provide no host-side support for converting between `cl_half` and other integer or floating-point types.

The following code snippet

```
cl_float myfloat = ...
cl_half myhalf = (cl_half)myfloat;
```
results in a floating-point to integer conversion. This is really not nice.

While we can't make the above work in applications written in C, I suggest we provide conversion routines like https://github.com/KhronosGroup/OpenCL-CTS/blob/master/test_common/harness/imageHelpers.cpp#L898 as part of the headers. For applications written in C++, we could make casts work automatically.

kpet:
We should probably also provide `isnan`, `isnormal`, `isfinite` and `isinf`.

jrprice:
+1 to this, and I can do the work to get the initial conversion functions into the headers.

The functions in the CTS make extensive use of the `MAKE_HEX_FLOAT` macros. They also use `fabsf`. My current thinking is to rework them a little to avoid depending on these two things (easy enough to perform the comparisons as hex integers). Does this sound OK, or is there a desire to keep the functions exactly as they are in the CTS?

kpet:
That's great! Sounds good to me. I don't think we need/want to keep the functions exactly as they are in the CTS. I thought that we'd actually change the CTS to use the functions provided by the headers. The overall thinking was:
- starting from the functions in the CTS guarantees that the behaviour of the headers functions match what implementations do today
- using the functions provided in the headers in the CTS guarantees that applications use what implementations are verified with so at least the behaviour is consistent
- the last step would be to add unit tests for the headers functions vs. the spec and deal with the outcome of that.

kpet:
@jrprice #60 has done what I originally had in mind here. I've created https://github.com/KhronosGroup/OpenCL-CTS/issues/870 so we don't forget to transition the CTS to these routines. Since you added tests as well under #60 I'm not sure there's anything left to do here other than creating new issues to cover making conversions automatic in C++ programs and consider adding all the `is*` routines. Did you have something else in mind?

jrprice:
> Did you have something else in mind?

No, the C++ sugar and `is*` were the only reasons I didn't close this issue.

kpet:
Thanks! I've created #102 and #103. Closing this issue.

PR #60, opened 2019-12-23
Add cl_half.h header

jrprice:
This is going to address #57 when complete. I'm posting it now to get early feedback on the approach before I go on to do the half->float direction and FP64. 

This introduces a function to convert from FP32 to FP16. I've tested this by brute-forcing all FP32 
 values for all four rounding modes, and comparing against the functions in the CTS. I've also run the `half` suite from CTS using this function instead.

I've ended up essentially rewriting the function from scratch rather than copying the CTS versions, which is why I wanted to check with the group before pushing forwards. This new version does not depend on the hex float macros used in CTS, or on `math.h`. Also, while testing this I discovered that the versions in CTS do not always work correctly if certain compiler optimizations are enabled (due to the use of floating point arithmetic). This new version solely uses bitwise integer operations, so does not have this issue.

I've also opted to merge all the rounding modes into a single function, to avoid duplicating much of the code as is done in CTS. There's an `enum` to select the rounding mode. We could provide separate suffixed functions as well if that's desirable (e.g. `cl_float_to_half_rte()`).

I do not have any strong feelings about the function names, so feel free to suggest something different if desired.


jrprice:
This now has the following three conversion routines:
- FP32 -> FP16
- FP16 -> FP32
- FP64 -> FP16

I've tested all three in the `half` CTS suite on an NVIDIA GPU. I've also brute-forced all inputs for the first two, comparing bitwise to the same routines from CTS.

CL/cl_half.h
```diff
@@ -0,0 +1,454 @@
...
+
+  // Add FP16 exponent bias
+  uint16_t h_exp = exp + CL_HALF_MAX_EXP - 1;
+
+  // Check for NaN / infinity
+  if (f_exp == 0xFF)
+  {
+    if (f_mant)
+    {
+      // NaN -> propagate mantissa and silence it
```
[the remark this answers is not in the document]
jrprice: It's not a requirement, I just did this for compatibility with the conversion routines used in CTS. A fixed payload should work fine, if that's the preference. One argument for keeping the payload might be that it gives some symmetry with the FP16->FP32 direction, where I think we do need to propagate it (according to the paragraph you quoted). That said, I don't have a strong opinion on this.

CL/cl_half.h
```diff
@@ -0,0 +1,454 @@
...
+    lsb_pos = -exp + (CL_FLT_MANT_DIG - 25);
+  }
+  else
+  {
+    // Normal -> just truncate mantissa
+    lsb_pos = CL_FLT_MANT_DIG - CL_HALF_MANT_DIG;
+  }
+
+  // Generate FP16 mantissa by shifting FP32 mantissa
+  uint16_t h_mant = f_mant >> lsb_pos;
```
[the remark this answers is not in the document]
jrprice: It's not exactly clear from the code, but it shouldn't be possible for `lsb_pos` to be >=32,  since CL_FLT_MANT_DIG is only 24.

CL/cl_half.h
```diff
@@ -0,0 +1,458 @@
...
+extern "C" {
+#endif
+
+
+/**
+ * Rounding mode used when converting to cl_half.
+ */
+typedef enum
+{
+  CL_HALF_RTE, // round to nearest even
```
bashbaug: These enum values look like standard API enums but aren't in the spec or XML file.  Is that OK?  Should we add them to the spec / XML file?  Or, should we choose a different prefix to differentiate them from API enums?

CL/cl_half.h
```diff
@@ -0,0 +1,458 @@
...
+
+/* Private utility macros. */
+#define CL_HALF_EXP_MASK 0x7C00
+#define CL_HALF_MAX_FINITE_MAG 0x7BFF
+
+
+/*
+ * Utility to deal with values that overflow when converting to half precision.
+ */
+static inline cl_half cl_half_handle_overflow(cl_half_rounding_mode rounding_mode,
```
bashbaug: Confirming: the functions in this file use an underscore-delimited "snake case" naming convention as opposed to the OpenCL API "camel case" convention.  Is that OK?  Similar to the comment above, perhaps we should consider a different prefix to differentiate from standard API functions, if nothing else?

CL/cl_half.h
```diff
@@ -0,0 +1,458 @@
...
+                                              uint16_t sign)
```
bashbaug: If we're going to use the stdint types in this file then I think this header should explicitly include stdint.h.  cl_platform.h currently includes it for non-Windows, but it is not included on Windows.  Alternatively, we could use OpenCL types instead (such as cl_ushort), but that could be confusing.

CL/cl_half.h
jrprice: Agree this could be confusing - I'm happy to use a different prefix. What about something like `CLX_` or `CLU_` (`U` for Utility)? Or `OCL`?

CL/cl_half.h
jrprice: So the choice for snake case was in part to differentiate from the API functions. A prefix is a good idea though if we can find one we're happy with. As above, we could have `clxFloatToHalf`, `cluFloatToHalf` or `oclFloatToHalf`, but I'm very much open to suggestions (since none of these look nice to me!).

jrprice:
The comment about Windows reminds me to mention that I have *not* tested this on Windows yet. It'd be great if somebody with a Windows platform is able to do this before we merge this.

CL/cl_half.h
bashbaug: Thinking about this a bit more, I think I'd prefer to use a different prefix first and foremost.  I don't have a strong opinion regarding snake case or camel case, since the different prefix alone will ensure no conflicts with existing or future APIs or typedefs or enums.

I'm fine with either `clu` / `CLU`, or `clx` / `CLX`, or a different prefix.  I suppose if I had to choose I'd go with `clu`, even given the old [Computing Language Utility](https://github.com/Computing-Language-Utility/CLU).

```c
    cl_float f = 1.0f;
    cl_half h = clu_float_to_half(f, CLU_HALF_RTE);
    f = clu_half_to_float(h);
```

We should consider a different prefix for the name of this file, too.

bashbaug:
> I have _not_ tested this on Windows yet.

Windows compiled but generated some warnings (tested VS2019):

```
CL/cl_half.h(266,59): warning C4293:  '<<': shift count negative or too big, undefined behavior
CL/cl_half.h(272,18): warning C4244:  'initializing': conversion from 'int64_t' to 'uint16_t', possible loss of data
CL/cl_half.h(280,23): warning C4244:  'initializing': conversion from 'uint64_t' to 'uint16_t', possible loss of data
CL/cl_half.h(316,43): warning C4293:  '<<': shift count negative or too big, undefined behavior
CL/cl_half.h(319,20): warning C4244:  '=': conversion from 'int64_t' to 'uint32_t', possible loss of data
CL/cl_half.h(328,19): warning C4244:  'initializing': conversion from 'uint64_t' to 'uint16_t', possible loss of data
CL/cl_half.h(331,42): warning C4334:  '<<': result of 32-bit shift implicitly converted to 64 bits (was 64-bit shift intended?)
```

I believe the shift warnings are because `unsigned long` is still 32-bits on Windows, and can be fixed by replacing `1ul` with `(uint64_t)1`.  The remainder can be fixed with explicit casts.

CL/cl_half.h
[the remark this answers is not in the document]
kpet: Compatibility with the CTS routines is a good enough reason to keep this behaviour. Re-reading 6.2.3, keeping the payload when converting to a wider format of the same radix and then back to the same narrower is required, which definitely requires to preserve these mantissa bits. Let's keep things as is.

CL/cl_half.h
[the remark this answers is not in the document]
kpet: Yup, a bit tricky to spot but it's all fine.

CL/cl_half.h
```diff
@@ -0,0 +1,460 @@
...
+ * Example usage:
+ *
+ *    #include <CL/cl_half.h>
+ *    ...
+ *    cl_half h = cl_float_to_half(0.5f, CL_HALF_RTE);
+ *    cl_float f = cl_half_to_float(h);
+ */
+
+#ifndef __CL_HALF_H
+#define __CL_HALF_H
```
kpet: Didn't spot that the first time: `__CL_HALF_H` is technically a reserved identifier...

kpet:
>I also benchmarked the first drop of this code against the routines in CTS and it was faster, FWIW. I don't think performance is a particular priority for this header however (IMO).

That's great to hear! Agree performance is not a primary concern.

>I've pulled out some common routines for handling overflow and underflow.

Thanks! Looks good.

On naming generally:
- Having said that I liked the function names, I'm now wondering whether we shouldn't prefix all of them with `cl_half` (or whatever we agree should be the prefix): `cl_half_to_float`, `cl_half_from_double` and `cl_half_from_float`.
- @bashbaug I understand the concern but I really think this is functionality that we should ultimately provide in the cl namespace. I think what I'm saying is that I'd rather we document whatever additions we make (a spec appendix?) rather than move them to a separate namespace. One of the things I had in mind for future changes was to provide a better definition for `cl_half` for C++ applications where conversions and operations do the right thing automatically. If it looks like the discussions wouldn't converge reasonably quickly though, it might be better to just change the namespace for now (agree `clu` is reasonable) so we get forward progress. On the case, I think using snake case for these is a plus as it makes it clear these are not API functions.

bashbaug:
> what I'm saying is that I'd rather we document whatever additions we make (a spec appendix?) rather than move them to a separate namespace.

This seems like a reasonable solution too, if we would prefer to keep the `cl` prefix.

CL/cl_half.h
jrprice: True, but it's consistent with the other headers (e.g. `__CL_PLATFORM_H`), and we use double underscore for other internally used macros (`__CL_HAS_ANON_STRUCT__`). I can change this if it's a concern though.

jrprice: Actually, I see that you used a header guard without a leading double-underscore for the ICD header, so that's enough of a precedent for me to be a good citizen and do the same here.

jrprice:
I've fixed the Windows warnings (thanks @bashbaug!), and renamed the functions to `cl_half_from_float()` as suggested by @kpet .

I've left things in the `cl` namespace and using snake-case for now until we have consensus on the desired convention for these functions. I'm leaning (slightly) away from snake-case now since it makes them look a bit like the type definitions we have in the API. If we document them in a spec appendix, I'd be happy with `clHalfToFloat(...)` and `clHalfFromFloat(...)`.

kpet:
We've got two approvals from two different implementers. This can't break existing functionatlity. This has been discussed several times in the working group; nobody opposed the change. I'm suggesting we just merge this, any objections?

bashbaug:
> any objections?

No objections, but do note that this file has the old license header and will need to be updated (see #76).

kpet:
@jrprice With updated license headers, this could go in straight away ;).

kpet:
@jrprice Ping! This just needs a change of license headers and then we can pop the champagne :).

jrprice:
Sorry for the radio silence. The reason I've been holding off on doing this is because I'm still not clear that this header is best suited for this repository as opposed to the utility library in the OpenCL SDK, and we haven't had a chance to discuss this yet. Now that the fact we're doing an SDK is public maybe we can discuss this here.

So, given that this is a header providing utility functions, shouldn't it be in the SDK's utility library instead? I'm not sure I understand @kpet's previous point that having these here means we could have automatic conversion for C++ - you'd still need to include an extra header (i.e. not the same one that provides the `cl_half` definition), so surely that's no different from providing the functionality in the SDK instead?

kpet:
I had a suspicion that was the case :). The argument was that we could _with further work_ do that provided the conversion routines live in the headers repo. There are a number of approaches that come to mind:
- Include `cl_half.h` from `cl_platform.h`, always or only for C++. Doesn't sound quite right but can be made fully backwards compatible.
- Remove the existing definition in `cl_platform.h` and include `cl_half.h` from `cl.h`. Might break some code that includes `cl_platform.h` directly.
- Any other?

If we're giving up on trying to do that, then I agree that these routines are definitely better placed in the SDK. The question of the type remains though: would we want to provide a better definition for `cl_half` in the SDK and make the one in headers opt-out? Would we give up altogether on trying to make the right conversions automatic?

jrprice:
The part that hadn't clicked for me was that a different definition of `cl_half` would be needed to provide these conversion functions, which is why it would need to be done here. I have no further objections :-)

I've updated the license and also added some test coverage (not at all comprehensive, but better than nothing).

kpet:
There's been agreement on this PR for a while, we now have tests and correct license headers. Merging.

PR #63, opened 2020-01-14
Add version guards to ICD declarations

jrprice:
This allows the `cl_icd.h` header to be included when `CL_TARGET_OPENCL_VERSION` is defined to something other than the most recent version.

The `#ifdefs` make it a bit messy. The APIs could be grouped by version to aid this. I didn't do this yet to keep things consistent with `cl.h` and also make this change easier to review.

### DIFF
--- a/CL/cl_icd.h
+++ b/CL/cl_icd.h
@@ -30,6 +30,9 @@
 #define OPENCL_CL_ICD_H
 
 #include <CL/cl.h>
+#include <CL/cl_egl.h>
+#include <CL/cl_ext.h>
+#include <CL/cl_gl.h>
 
 #ifdef __cplusplus
 extern "C" {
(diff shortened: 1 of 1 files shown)


bashbaug:
Just FYI, I've been experimenting with generation of OpenCL headers and other files from the `cl.xml` file, and I've been able to generate a version of this `cl_icd.h` file, including version guards.  I've uploaded an example [here](https://gist.github.com/bashbaug/021b1770e6bf923a9410c9c395c1fb26).

It's not quite ready for prime time, so it definitely shouldn't hold up these changes, but I did want to pass this along in case somebody else is thinking of doing something similar.

kpet:
@bashbaug That's great! Looking forward to seeing `cl.xml` used more. I've started using it for a couple of things as well.

Merging this.

oddhack:
I got pretty far along with generating CL headers from the XML when doing the asciidoctor conversion, but there were so many hand-coded odd things in the core headers that it looked like a pretty steep task to replicate them functionally exactly (let alone layout exactly, and some people feel extremely strongly about such things). Missing demand from the WG I just let it sit, but if there's appetite to revisit that I can try to dig it up again - would really like to keep schema and scripts lined up with Vulkan and OpenXR. Probably the most complex thing to handle at the time was all the DEPRECATION macros - I put attributes into the XML (prefix= and suffix=) to capture their existence, but the header generation has no idea what to do with those attributes.

Issue #58, opened 2019-12-13
Move all the headers CTS tests to this repository?


kpet:
There are a number of so-called "headers" tests in the CTS (https://github.com/KhronosGroup/OpenCL-CTS/tree/master/test_conformance/headers) that don't really test the OpenCL implementation itself but rather whether the headers build cleanly using the target toolchain and whether some aspects of them behave correctly.

I think these tests should be moved to the headers repo as they're effectively testing the headers in isolation. However, I still see value in requiring that these tests pass on a platform for which OpenCL compliance is sought. Maybe this can be a requirement of the conformance process?

Any opinions?

kpet:
This happened a long time ago. Closing.

PR #64, opened 2020-01-15
Add tests and Travis CI config

jrprice:
The tests are copied from OpenCL-CTS. A small amount of additional CMake work will be needed to test on Windows as well.

Contributes to #58.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,16 @@
+language: c
+
+compiler:
+  - gcc
+  - clang
+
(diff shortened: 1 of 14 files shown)


jrprice:
Addressed comment about system include path and added more tests. The only headers that aren't covered are the d3d/dx ones (I'll let someone else tackle Windows), and the `cl_va_api_media_sharing_intel.h` (which has an external dependency).

Issue #66, opened 2020-01-28
License mismatch


jrprice:
I've just noticed that this repository is using an MIT license, whereas most of our others use Apache 2.0. The new test files that I added recently (copied from CTS) use Apache 2.0, which doesn't match the MIT license specified in the root `LICENSE` file in this repository.

IANAL, so I don't know if this matters and/or what we should do about it, so opening this issue just to check with others. Can we change either the licenses for the new tests or the licenses for the headers themselves? Should we just add the Apache 2.0 license text to the root `LICENSE` file? Or do nothing since the test sources were originally released under Apache 2.0 before being copied here?

alycm:
IANAL also, but "do nothing" doesn't seem acceptable.

Documenting the additional code in the root LICENSE file would be the minimum.  I expect that would be sufficient for myself, but given how widely the headers are shipped it is possible that others would have stricter requirements.

AFAIK the fact that it is all Khronos copyright means that relicensing (or dual licensing) the copied code is an option (with approval).

PR #76, opened 2020-03-17
Change license to Apache 2.0

jrprice:
As agreed by the OpenCL WG on the 2020-03-17 teleconference.

I've preserved the additional copyright notices in the Intel extension headers where applicable. I've also removed some old SVN revision tags that are no longer relevant.

Fixes KhronosGroup/OpenCL-Headers#66.

### DIFF
--- a/CL/cl.h
+++ b/CL/cl.h
@@ -1,29 +1,17 @@
 /*******************************************************************************
- * Copyright (c) 2008-2019 The Khronos Group Inc.
+ * Copyright (c) 2008-2020 The Khronos Group Inc.
  *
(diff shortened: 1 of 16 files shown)


bashbaug:
This mostly looks good to me.  My one slight theoretical concern is that the license headers are now using C++-style comments (`//`) rather than C-style comments (`/* */`).  Since these are expected to be C headers, should we use C-style comments instead?

jrprice:
Fair point, I've switched them to C-style comments.

bashbaug:
Thanks!  Merging as discussed in the March 24th teleconference.

Issue #80, opened 2020-03-31
Add Windows CI and fix errors


kpet:
At least `cl_icd.h` has got issues on Windows (it's not self-contained). We should add CI on Windows and fix existing errors. Extending the existing Travis CI config should work.

PR #81, opened 2020-03-31
Make CL/cl_icd.h self-contained

fodinabor:
Update included DirectX 10 header and include the DirectX sharing headers from CL/cl_icd.h.
This addresses #80 and is necessary to build clvk on Windows.

### DIFF
--- a/CL/cl_icd.h
+++ b/CL/cl_icd.h
@@ -34,6 +34,12 @@
 #include <CL/cl_ext.h>
 #include <CL/cl_gl.h>
 
+#if defined(_WIN32)
+#include <CL/cl_d3d11.h>
+#include <CL/cl_d3d10.h>
+#include <CL/cl_dx9_media_sharing.h>
+#endif
+
 #ifdef __cplusplus
 extern "C" {
 #endif


fodinabor:
I did some research on this and looking at the [documentation](https://docs.microsoft.com/en-us/windows/win32/api/d3d10_1/) of the interfaces in `d3d10_1.h`. The notes that all those interfaces are available since Windows Vista SP1 makes me suggest to test which Windows SDK is targeted, like this:

```c++
#include <sdkddkver.h>
#if WDK_NTDDI_VERSION >= 0x06000100 // NTDDI_VISTASP1
#include <d3d10_1.h>
#else
#include <d3d10.h>
#endif
```

kpet:
That looks reasonable to me. Maybe there's a case for splitting this into two PRs. The `cl_icd.h` changes can go in right now. The others will need feedback from folks who work with Windows more regularly than I do (@bashbaug @nikhiljnv may have an opinion).

fodinabor:
Ok I investigated the include problem a bit more and I realized that it was a mere issue of the ordering of the `CL/cl_d3d11.h` and `CL/cl_d3d10.h` headers.
If both `d3d10.h` and `d3d10_1.h` are included, `d3d10_1.h` has to be included first.

And as `CL/cl_d3d11.h`  includes the `d3d10_1.h` transitively, it must be included before `d3d10.h` is included in `cl_d3d10.h`.

Therefore the current version in this PR should be enough to make `CL/cl_icd.h` self-contained and  enable the building on windows correctly.

PR #82, opened 2020-04-04
Add Windows CI using Travis

### LINKED ISSUE #80: Add Windows CI and fix errors

kpet:
Fixes #80

Signed-off-by: Kévin Petit <kpet@free.fr>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,9 +8,13 @@ os:
   - linux
   - osx
 
+jobs:
+  include:
+  - os: windows
+
 script:
   - mkdir -p build
   - cd build
   - cmake ../tests
-  - make
-  - make test
+  - cmake --build .
+  - ctest -C Debug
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -8,9 +8,13 @@ os:
   - linux
   - osx
 
+jobs:
+  include:
+  - os: windows
```
bashbaug, line 13: Is this a known Travis pattern?  I was expecting to see "windows" in the OS list above, similar to "linux" and "osx".  The pattern here works, and it seems to be better in some ways since it doesn't spawn separate "clang" and "gcc" jobs for Windows (even though Windows appears to be exclusively using Visual Studio 2017 - I think).  Since it's a bit unconventional I'd consider adding a few comments to explain what's going on, though.

kpet: Yes, it's fairly typical with Travis and well documented at https://docs.travis-ci.com/user/build-matrix/. I can add a comment pointing to the documentation if you want.

bashbaug: TIL... thanks!  No need to add a comment, I think.

PR #88, opened 2020-04-27
update headers for OpenCL 3.0

bashbaug:
Updates headers for the OpenCL 3.0 provisional spec.


bashbaug:
Notes:

1. I didn't change the default OpenCL version for the headers, since OpenCL 3.0 is still a provisional spec.  So, developers will need to explicitly opt-into OpenCL 3.0 by defining `CL_TARGET_OPENCL_VERSION` as `300`, otherwise the default remains OpenCL 2.2.

2. I switched from the `CL_EXPERIMENTAL` define to `CL_VERSION_3_0`.

3. I didn't put an `#ifdef` around the new APIs in the the ICD dispatch table struct.  So, they'll always exist in the dispatch table, but they may be `void*`s for pre-OpenCL 3.0 headers.

I'm pretty sure (1) is the right thing to do but I'm less sure about (2) and (3).

Requesting a review from @alycm, to ensure these changes are compatible with the ICD loader changes in https://github.com/KhronosGroup/OpenCL-ICD-Loader/pull/108.  Thanks!

alycm:
(1) Makes sense.

(2) Also makes sense to me and will make the transition from provisional to final smooth for anyone experimenting and implementers.

(3) This is very low risk and I have no objection to it.  The gain from it I suppose is that tools that are ICD loader aware would have the same smooth transition benefit that I noted for (2).

bashbaug:
Thanks for the speedy review!

I've updated the tests to include OpenCL 3.0 testing, which I had missed previously.

I'll plan to merge these changes first thing tomorrow unless there are other review comments.

bashbaug:
Thanks for the reviews!  Merging.

PR #93, opened 2020-06-19
Fix build of dependent software with clang

pkubaj:
When using clang, altivec.h inclusion is taken care of by clang with -maltivec / -mno-altivec. Users shouldn't directly include it, it breaks compilation.

### DIFF
--- a/CL/cl_platform.h
+++ b/CL/cl_platform.h
@@ -355,7 +355,9 @@ typedef unsigned int cl_GLenum;
 
 /* Define basic vector types */
 #if defined( __VEC__ )
-   #include <altivec.h>   /* may be omitted depending on compiler. AltiVec spec provides no way to detect whether the header is required. */
+  #if !defined(__clang__)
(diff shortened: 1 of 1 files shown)


bashbaug:
I'm fine merging this, but I do have one basic question (note, I am not an altivec expert!): Was including altivec.h ever valid for Clang?  In other words, do we need a more precise `#ifdef` than just `__clang__`, or is it fine to skip including altivec.h for all Clang versions?

Can we enhance our CI testing to catch this issue?

Thanks!

pkubaj:
I don't think so, I have LLVM 10.0.1 right now, but I remember fixing other software that did the same back when I was using LLVM 8.0. 8.0 was the first version that was usable on ppc targets. Previous versions didn't support C++ exceptions.

bashbaug:
Thanks for the prompt reply.  I'll merge this change - it's clearly broken now and we can always add improvements later, if necessary.

Issue #86, opened 2020-04-22
Policy for Extension Preprocessor Defines


bashbaug:
The policy for extension preprocessor defines - meaning a `#define` for `cl_my_extension_name` in the headers - is currently a bit ad-hoc.  It looks like the original intent was that an extension should have a define if it adds new API functions, but not otherwise, e.g.:

```c
/* cl_khr_fp64 extension - no extension #define since it has no functions  */
```

But, some extensions include a define anyways, either because they erroneously included a define initially and this has been carried forward for backwards compatibility, or because an extension define can be useful even when it doesn't add new API functions - see #85 and https://github.com/KhronosGroup/OpenCL-CLHPP/pull/82.

Should we just add an extension define in all cases, uniformly?

PR #104, opened 2020-08-06
Align headers with cl_intel_create_buffer_with_properties and cl_intel_mem_channel_property extensions

vmaksimo:
- Add cl_intel_create_buffer_with_properties extension function
Spec: https://github.com/KhronosGroup/OpenCL-Docs/blob/master/extensions/cl_intel_create_buffer_with_properties.asciidoc
This extension allows OpenCL 1.x and 2.x devices to support the clCreateBufferWithProperties API that was added in OpenCL 3.0. This allows older OpenCL implementations to support other optional extensions or features that use the clCreateBufferWithProperties API to specify additional buffer properties, without recreating the API that is already part of OpenCL 3.0.

- Add CL_MEM_CHANNEL_INTEL property enum
Spec: https://github.com/KhronosGroup/OpenCL-Docs/blob/master/extensions/cl_intel_mem_channel_property.asciidoc
This extension allows programmers to request that a buffer allocation be implemented in a particular region of memory.

### DIFF
--- a/CL/cl_ext_intel.h
+++ b/CL/cl_ext_intel.h
@@ -628,6 +628,24 @@ clEnqueueMemcpyINTEL_fn)(
             const cl_event* event_wait_list,
             cl_event* event);
 
+extern CL_API_ENTRY cl_mem CL_API_CALL
+clCreateBufferWithPropertiesINTEL(
+    cl_context   context,
+    const cl_mem_properties_intel* properties,
+    cl_mem_flags flags,
+    size_t       size,
+    void *       host_ptr,
+    cl_int *     errcode_ret) CL_EXT_SUFFIX__VERSION_1_0;
+
+typedef CL_API_ENTRY cl_mem (CL_API_CALL *
+clCreateBufferWithPropertiesINTEL_fn)(
+    cl_context   context,
+    const cl_mem_properties_intel* properties,
+    cl_mem_flags flags,
+    size_t       size,
+    void *       host_ptr,
+    cl_int *     errcode_ret) CL_EXT_SUFFIX__VERSION_1_0;
+
 #ifdef CL_VERSION_1_2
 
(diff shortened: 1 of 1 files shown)


mlychkov:
@bashbaug Could you please review and merge this PR if there is no concerns?

CL/cl_ext_intel.h
```diff
@@ -628,6 +629,24 @@ clEnqueueMemcpyINTEL_fn)(
             const cl_event* event_wait_list,
             cl_event* event);
 
+extern CL_API_ENTRY cl_mem CL_API_CALL
```
bashbaug, line 632: Can you please distinguish these extension functions from the USM extension functions?  Right now they're intermingled with the USM extension functions, which is a little confusing.

Suggestions:

* Move these two functions to the end of this file.
* Add a header comment similar to the other extensions in this file:

```c
/*****************************************
* cl_intel_create_buffer_with_properties *
******************************************/
```

CL/cl_ext_intel.h
```diff
@@ -628,6 +629,24 @@ clEnqueueMemcpyINTEL_fn)(
...
+clCreateBufferWithPropertiesINTEL(
```
bashbaug, line 633: Since this extension adds new APIs (and not just new enums, though see #86), it should add an extension `#define`:

```c
#define cl_intel_create_buffer_with_properties 1
```

mlychkov:
> Thanks, this mostly looks good, just a few things to fix.
> 
> Could you please also create a PR to add this extension to the XML grammar?
> 
> https://github.com/KhronosGroup/OpenCL-Docs/blob/master/xml/cl.xml

Thank you very much for such detailed feedback.

vmaksimo:
Resolved and merged in https://github.com/KhronosGroup/OpenCL-Headers/pull/106

Issue #94, opened 2020-07-14
OpenCL-Header versioning


nikhiljnv:
Currently OpenCL headers are versioned based on the OpenCL spec (1.2, 2.0, 2.2 etc)

Apart from the core and optional core features that we change only with OpenCL version, we often add types and APIs required by KHR extensions in some of the standard headers (e.g. clext.h). Sometimes, the extension might be written on top of 1.2 or 2.0.

Current versioning doesn't allow differentiating between the OpenCL headers before and after such changes. Having finer versioning may help recommend right version of headers users need to use and potentially clarify this dependency in the extension specs. 

Filing this issue to discuss pros and cons of adding the finer versioning.

kpet:
Linking https://github.com/KhronosGroup/OpenCL-Headers/issues/56

PR #107, opened 2020-08-11
Comprehensive CMake Package Config support

MathiasMagnus:
# Summary
This PR is part of a set of PRs implementing CMake Package Config detection support for all OpenCL repos independently and as part of the OpenCL SDK as well.

Related PRs:
- [OpenCL-CLHPP](https://github.com/KhronosGroup/OpenCL-CLHPP/pull/109)
- [OpenCL-ICD-Loader](https://github.com/KhronosGroup/OpenCL-ICD-Loader/pull/115)

The PR also implements Github Actions, as it is required to thoroughly test that the new CMake logic does not conflict with existing practice. I tried keeping attribution where applicable with parallel PRs were in place. I still have little knowledge of Conan (on my short list) but I kept that part in good faith as all being mandatory to have Conan support.

## Package Config support
CMake has two dependency detection mechanisms:

- Find Modules
- Package Configs

The first is what CMake currently offers and typically caters to projects that aren't CMake aware. Find Modules look in a few pre-determined places and try to detect typical installations. The more modern and robust dependency detection mechanism are Package Config scripts, which are typically produced by CMake builds themselves (though not necessarily) and describe the installation fully. Finding a Package Config file ensures that detection has succeeded fully. These two methods can coexist side-by-side.

The idea is that users could rely on the existing Find Module detection, but also the Package Config detection if they so wished. CMake caters to such scenarios through the two `find_package()` signatures ([full docs here](https://cmake.org/cmake/help/latest/command/find_package.html)):

| Basic Signature and Module Mode | Full Signature and Config Mode |
| --------------- | --------------- |
| `find_package(<PackageName> [LONG_LIST_OF_COMMON_ARGS] [MODULE] [NO_POLICY_SCOPE])` | `find_package(<PackageName> [LONG_LIST_OF_COMMON_ARGS] [MODULE] [CONFIG\|NO_MODULE] [WHOOOOLE_BUNCH_OF_EXTRA_ARGS]` |

CMake when using the 'old school' basic signature (like all existing OpenCL detection) will favor Module Mode if both detections were to succeed. Without modifying the scripts, users can still opt-in to Config Mode from the command-line via [`CMAKE_FIND_PACKAGE_PREFER_CONFIG`](https://cmake.org/cmake/help/latest/variable/CMAKE_FIND_PACKAGE_PREFER_CONFIG.html#variable:CMAKE_FIND_PACKAGE_PREFER_CONFIG)

From the infinite dimension phase space, here's a possible script disregarding Module Mode:

```cmake
cmake_minimum_required(VERSION 3.0)

project(PkgConfigTest
  LANGUAGES C
)

find_package(OpenCL
  REQUIRED
  CONFIG
  COMPONENTS Headers
)

add_executable(${PROJECT_NAME}
  pkgconfig.c
)

target_link_libraries(${PROJECT_NAME}
  OpenCL::Headers
)

target_compile_definitions(${PROJECT_NAME}
  PRIVATE
    CL_TARGET_OPENCL_VERSION=120
)
```

The idea is that existing, untouched scripts calling

```
find_package(OpenCL REQUIRED)
...
target_link_libraries(${PROJECT_NAME} OpenCL::OpenCL)
```

would continue to work, but will require for all the PRs with package config to go through. The list of imported targets projects will create are:

- `OpenCL::Headers` here and now
- `OpenCL::CppHeaders` from the CLHPP repo
- `OpenCL::ICD` from the ICD repo
- `OpenCL::OpenCL` an alias to the three above, registered by the 'super project', the SDK
- `OpenCL::SDK` the SDK utility library, also defined by the SDK

Individual components could not be depended on by the older, Find Module mode detection, but that shouldn't be a problem, as such consumers would expect all components to be available, should detection succeed. This scenario will expand on the additional `OpenCLConfig.cmake` which the SDK will install and also allow shiny new detection and old-school consumption using `${OpenCL_INCLUDE_DIRS}` et al. such as:

```
find_package(OpenCL REQUIRED)
add_executable(${PROJECT_NAME} Main.cpp)
target_include_directories(${PROJECT_NAME} PRIVATE ${OpenCL_INCLUDE_DIRS})
target_link_libraries(${PROJECT_NAME} PRIVATE ${OpenCL_LIBRARIES})
```

## GitHub Actions
Tapping into the resources provided by Microsoft to open-source projects for free, I've implemented CI using GitHub Actions. This new CI definition greatly expands on compiler and use-case coverage over the current Jenkins-based solution. There are still scenarios missing (cross-compilation most importantly), but since we don't test those today, it's still more than what we've got now. Adding those could be iterative additions to this code drop.

The test matrices are defined as can be seen from the CI itself:

```yaml
jobs:
  linux-gcc:
    runs-on: ubuntu-latest
    strategy:
      matrix:
        VER: [7, 8, 9]
        EXT: [ON, OFF]
        GEN: [Unix Makefiles, Ninja]
        VAR: [Debug, Release]
        STD: [11] # 90 results in errors
        BIN: [32, 64]

  windows-msvc:
    runs-on: windows-latest
    strategy:
      matrix:
        VER: [v140, v141, v142]
        EXT: [ON, OFF]
        GEN: [Visual Studio 16 2019, Ninja Multi-Config]
        BIN: [x64, x86]
        #STD Proper 11, 17 incoming: https://devblogs.microsoft.com/cppblog/c11-and-c17-standard-support-arriving-in-msvc/

  macos-gcc:
    runs-on: macos-latest
    strategy:
      matrix:
        VER: [8, 9]
        EXT: [ON, OFF]
        GEN: [Xcode, Ninja Multi-Config]
        STD: [11] # 90 results in errors
```

### Notes:
- As the primary motivation for the CI rework was the `/Za` fiasco with MSVC, I'm happy to announce GCC too rejects the headers in strict ANSI (C90) conformance mode, hence this language mode is currently disabled in the CI. Whether we want to enable and fix is up for future discussion.
- `EXT` refers to compiler/platform-specific language extensions. Effectively controlls `-std=gnu++11` vs. `-std=c++11`. See docs [here](https://cmake.org/cmake/help/latest/variable/CMAKE_CXX_EXTENSIONS.html).
- MSVC currently doesn't support selecting any language version other than ANSI. (Non-ANSI mode is an arbitrary mix of language versions and extensions.) Proper C11 & C17 mode will shortly be available when using the v142 toolset.
- We're using the latest MSBuild schema but test with multiple compilers back to VS 2015 (toolset v140).
- We test with GCC on MacOS, however there is some accidental AppleClang testing too, as XCode prefers that over GCC, so in reality the XCode generator uses latest AppleClang compilers as opposed to GCC which is provided on the command line. If anyone know the secret sauce of how to tell CMake to generate an XCode project with the correct toolset parameter, let me know.
- There are no Clang tests yet, but we don't do those today either. Should be added later.
- I do intend to rework the _massive_ amount of script with a fair amount of code duplication, but didn't want to hold up the feature any longer.
  - Ultimately I want to rewrite the YAML bash/cmd snippets as portable CMake scripts which are invoked in script mode as `cmake ... -P build.cmake` The YAML would only be used to define the test matrix and the number of steps would reduce as well as become script one-liners. Complexity would be offloaded to one script.
  - The _insane_ benefit of this would be to omit having to run GitHub Actions / GitLab Runners locally, but one could invoke the same one-liner in their local docker installs on their devbox and see how CI would perform, but locally. Once YAML is only used to define the test matrix and nothing else, one could easily 're-implement' that to run whatever they want to / can test locally; just a shell script with 2-3 nested loops enumerating a subset of the matrix and invoking `build.cmake` as does the CI runner.


bashbaug (commented):
Thanks for putting this together!

I've been trying to test these changes, so far without luck.  Is there a dependency on some of the other PRs you've linked?  Here is what I'm doing:

I've built  the package as described in this PR.  This seems to be working properly:

```sh
$ cmake -S . -B build -DCMAKE_INSTALL_PREFIX=/home/bashbaug/git/header_install
-- The C compiler identification is GNU 7.5.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Configuring done
-- Generating done
-- Build files have been written to: /home/bashbaug/git/SimpleOpenCLSamples/external/OpenCL-Headers/build

$ cmake --build build --target install
Install the project...
-- Install configuration: ""
-- Up-to-date: /home/bashbaug/git/header_install/include/CL
-- Up-to-date: /home/bashbaug/git/header_install/include/CL/cl_platform.h
<snip>
-- Up-to-date: /home/bashbaug/git/header_install/include/CL/cl_d3d11.h
-- Installing: /home/bashbaug/git/header_install/share/cmake/OpenCLHeaders/OpenCLHeadersTargets.cmake
-- Installing: /home/bashbaug/git/header_install/share/cmake/OpenCLHeaders/OpenCLHeadersConfig.cmake
-- Installing: /home/bashbaug/git/header_install/share/cmake/OpenCLHeaders/OpenCLHeadersConfigVersion.cmake
```

I have a test app to use the package:

```sh
$ ls
CMakeLists.txt  main.cpp

$ cat CMakeLists.txt 
cmake_minimum_required(VERSION 3.1)
project(test)
add_executable(test main.cpp)
find_package(OpenCLHeaders REQUIRED)
target_link_libraries(test PRIVATE OpenCL::OpenCLHeaders)
```

I'm trying to build it with the package, but I'm getting errors:

```sh
$ cmake -S . -B build -DOpenCLHeaders_ROOT=/home/bashbaug/git/header_install
-- The C compiler identification is GNU 7.5.0
-- The CXX compiler identification is GNU 7.5.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
CMake Warning (dev) at CMakeLists.txt:4 (find_package):
  Policy CMP0074 is not set: find_package uses <PackageName>_ROOT variables.
  Run "cmake --help-policy CMP0074" for policy details.  Use the cmake_policy
  command to set the policy and suppress this warning.

  CMake variable OpenCLHeaders_ROOT is set to:

    /home/bashbaug/git/header_install

  For compatibility, CMake is ignoring the variable.
This warning is for project developers.  Use -Wno-dev to suppress it.

CMake Error at CMakeLists.txt:4 (find_package):
  By not providing "FindOpenCLHeaders.cmake" in CMAKE_MODULE_PATH this
  project has asked CMake to find a package configuration file provided by
  "OpenCLHeaders", but CMake did not find one.

  Could not find a package configuration file provided by "OpenCLHeaders"
  with any of the following names:

    OpenCLHeadersConfig.cmake
    openclheaders-config.cmake

  Add the installation prefix of "OpenCLHeaders" to CMAKE_PREFIX_PATH or set
  "OpenCLHeaders_DIR" to a directory containing one of the above files.  If
  "OpenCLHeaders" provides a separate development package or SDK, be sure it
  has been installed.


-- Configuring incomplete, errors occurred!
See also "/home/bashbaug/git/SimpleOpenCLSamples/external/OpenCL-Headers/my_test_app/build/CMakeFiles/CMakeOutput.log".
```

I can get a bit farther with a longer path and `-DOpenCLHeaders_DIR` vs. `-DOpenCLHeaders_ROOT`, but this produces a different error:

```sh
$ cmake -S . -B build -DOpenCLHeaders_DIR=/home/bashbaug/git/header_install/share/cmake/OpenCLHeaders/
-- The C compiler identification is GNU 7.5.0
-- The CXX compiler identification is GNU 7.5.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Configuring done
CMake Error at CMakeLists.txt:3 (add_executable):
  Target "test" links to target "OpenCL::OpenCLHeaders" but the target was
  not found.  Perhaps a find_package() call is missing for an IMPORTED
  target, or an ALIAS target is missing?


-- Generating done
CMake Generate step failed.  Build files cannot be regenerated correctly.
```

I'm pretty sure this is a problem on my end, but I would like to see this working before approving.  

A bit more of my system info in case it's helpful:

```sh
$ uname -a
Linux bashbaug-nuc 5.4.0-47-generic #51~18.04.1-Ubuntu SMP Sat Sep 5 14:35:50 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
$ cmake --version
cmake version 3.18.2

CMake suite maintained and supported by Kitware (kitware.com/cmake).
```

Thanks!

MathiasMagnus:
@bashbaug The why you get the error is because the name of the imported target is `OpenCL::Headers`. I wanted to omit meaningless word duplication. (Updating initial comment now, as feature is deemed production ready.)

.github/workflows/cmake.yml
```diff
@@ -0,0 +1,360 @@
...
+        if [[ "${{matrix.BIN}}" == "32" ]]; then sudo apt install -y gcc-multilib gcc-8-multilib gcc-9-multilib; fi
+      # Install Ninja only if it's the selected generator and it's not available.
+
+    - name: Configure CMake
+      shell: bash
+      run: cmake
+        -G "${{matrix.GEN}}"
+        -D BUILD_TESTING=ON
+        -D CMAKE_BUILD_TYPE=${{matrix.VAR}}
+        -D CMAKE_C_FLAGS="-Wall -Wextra -Werror -pedantic -Wno-format -m${{matrix.BIN}}"
```
kpet, line 35: We'll probably want to fix format warnings eventually. Would be good to have an issue for this.

.github/workflows/cmake.yml
```diff
@@ -0,0 +1,360 @@
...
+jobs:
+  linux-gcc:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        VER: [7, 8, 9]
+        EXT: [ON, OFF]
+        GEN: [Unix Makefiles, Ninja]
+        VAR: [Debug, Release]
+        STD: [11] # 90 results in errors
```
kpet, line 14: What about 99? We should probably test C++ as well. I'm not saying this should be in scope for this PR, just that we should track/discuss it.

CMakeLists.txt
```diff
@@ -0,0 +1,66 @@
+cmake_minimum_required(VERSION 3.0)
+
+project(OpenCLHeaders
+  VERSION 2.2
```
kpet, line 4: Should this be 3.0? We're using a date scheme for tags. What should we use here?

CMakeLists.txt
```diff
@@ -0,0 +1,66 @@
...
+  FILE OpenCLHeadersTargets.cmake
+  NAMESPACE OpenCL::
+  DESTINATION ${config_package_location}
+)
+install(
+  FILES ${CMAKE_CURRENT_BINARY_DIR}/OpenCLHeaders/OpenCLHeadersConfig.cmake
+  DESTINATION ${config_package_location}
+)
+
+unset(CMAKE_SIZEOF_VOID_P)
```
kpet, line 51: Why?

tests/pkgconfig/sdk/pkgconfig.c
```diff
@@ -0,0 +1,6 @@
+#include <CL/cl.h>
+
+int main()
+{
+    return sizeof(cl_platform_id) - sizeof(cl_context);
+}
```
kpet, line 6: This is the same file as ` tests/pkgconfig/bare/pkgconfig.c`. Do we need two copies?

tests/test_cl_half.h.c
```diff
@@ -95,7 +95,7 @@ int main(void)
   CHECK_FROM_FLOAT(-(CL_HALF_MAX + 1000.f), 0xfbff, RTP);
   CHECK_FROM_FLOAT(-(CL_HALF_MAX + 1000.f), 0xfc00, RTN);
   CHECK_FROM_FLOAT(-(CL_HALF_MAX + 1000.f), 0xfbff, RTZ);
-
+#ifndef _MSC_VER
```
kpet, line 98: How are we dealing with this in the CTS? Maybe we could wrap hex float constants in macros and make this transparent (just thinking out loud).

kpet (commented):
I've just had a quick skim but this is an awesome step forward.

.github/workflows/cmake.yml
MathiasMagnus: #132 written out.

.github/workflows/cmake.yml
MathiasMagnus: Added.

CMakeLists.txt
MathiasMagnus: Because we got shapeshifting headers based on defines, I see two meaningful choices:
- Maximum supported version
- Date tag

The question to ask here: what would users want to test for? The availability of an API version in the headers, or a fix to some bug? Both checks are likely made in build systems. IMHO we should offload this question to a separate issue, as we've already have #94 as an anchor for this. I vote for a separate PR, a oneliner even.

CMakeLists.txt
MathiasMagnus: I took this from the the previous contributor @sl1pnk07, I believe this is in order to have packages always install to `lib` and not `lib64` on 64-bit platforms. I'm not entirely sure. If this is the reason, it would cause problem only if people installed their own builds of something into system paths. That is a slippery slope and generally calling for trouble.

tests/pkgconfig/sdk/pkgconfig.c
MathiasMagnus: No, not really. It depends on what triggers someone: code duplication of a "one-liner" or referencing source files outside the project dir. You may be right, that code duplication can more easily bite. Fixed.

tests/test_cl_half.h.c
MathiasMagnus: I don't see host-side tests for cl_half, only those which load-store halfs from various memory spaces and the combination, doing a roundtrip to device and back. There are no rounding tests. These on MSVC are hard, syntax errors. Opened #133 to defer resolution.

.github/workflows/cmake.yml
kpet: Thanks!

.github/workflows/cmake.yml
kpet: That's good enough for a first step, resolving.

tests/pkgconfig/sdk/pkgconfig.c
kpet: Yeah, agree it was not a lot of duplication. Thanks for removing it anyway.

tests/test_cl_half.h.c
kpet: Fine with deferring to a follow on issue.

CMakeLists.txt
kpet: I think we should leave a TODO or create an issue if we don't really understand this fully.

CMakeLists.txt
kpet: Yeah, agree this needs a bit of discussion. I think using the maximum supported version is the most likely outcome. We could bump this to 3.0 and defer the full discussion to a separate issue.

Issue #115, opened 2020-09-21
MSVC preprocessor codepaths mistreat strict ANSI conformance (/Za)


MathiasMagnus:
When compiling an empty `main` including `cl.h` the compiler bails out on anonymous structs when compiling in strict ANSI conformance using the [/Za](https://docs.microsoft.com/en-us/cpp/build/reference/za-ze-disable-language-extensions?view=vs-2019) flag. This compiler option also [defines `__STDC__`](https://docs.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=vs-2019) which can be used to detect such use cases and be added to [L487 of `cl_platform.h`](https://github.com/KhronosGroup/OpenCL-Headers/blob/master/CL/cl_platform.h#L487) to read:
```
#elif defined( _WIN32) && defined(_MSC_VER) && ! defined(__STDC__)
```
This issue is here for me to remember ro open a PR and what the issue is (since I got to the bottom of it).

PR #116, opened 2020-09-21
Avoid anon structs when MSVC uses /Za

MathiasMagnus:
This PR addresses #115

### DIFF
--- a/CL/cl_gl_ext.h
+++ b/CL/cl_gl_ext.h
@@ -30,7 +30,7 @@ extern "C" {
 
 extern CL_API_ENTRY cl_event CL_API_CALL
 clCreateEventFromGLsyncKHR(cl_context context,
-                           cl_GLsync  /*cl_GLsync*/,
+                           cl_GLsync  sync,
                            cl_int *   errcode_ret) CL_EXT_SUFFIX__VERSION_1_1;
 
 #ifdef __cplusplus
--- a/CL/cl_gl_ext.h
+++ b/CL/cl_gl_ext.h
@@ -30,7 +30,7 @@ extern "C" {
 
 extern CL_API_ENTRY cl_event CL_API_CALL
 clCreateEventFromGLsyncKHR(cl_context context,
-                           cl_GLsync  cl_GLsync,
+                           cl_GLsync  /*cl_GLsync*/,
                            cl_int *   errcode_ret) CL_EXT_SUFFIX__VERSION_1_1;
 
 #ifdef __cplusplus
--- a/CL/cl.h
+++ b/CL/cl.h
@@ -133,22 +133,24 @@ typedef struct _cl_image_desc {
     size_t                  image_slice_pitch;
     cl_uint                 num_mip_levels;
     cl_uint                 num_samples;
-#ifdef CL_VERSION_2_0
-#ifdef __GNUC__
-    __extension__   /* Prevents warnings about anonymous union in -pedantic builds */
-#endif
+#ifdef CL_VERSION_2_0 && __CL_HAS_ANON_STRUCT__
 #ifdef _MSC_VER
+#if _MSC_VER >= 1500
 #pragma warning( push )
 #pragma warning( disable : 4201 ) /* Prevents warning about nameless struct/union in /W4 /Za builds */
 #endif
+#endif
+    __CL_ANON_STRUCT__
     union {
 #endif
       cl_mem                  buffer;
-#ifdef CL_VERSION_2_0
+#ifdef CL_VERSION_2_0 && __CL_HAS_ANON_STRUCT__
       cl_mem                  mem_object;
     };
 #ifdef _MSC_VER
+#if _MSC_VER >= 1500
 #pragma warning( pop )
+#endif
 #endif
 #endif
 } cl_image_desc;
--- a/CL/cl.h
+++ b/CL/cl.h
@@ -133,7 +133,8 @@ typedef struct _cl_image_desc {
     size_t                  image_slice_pitch;
     cl_uint                 num_mip_levels;
     cl_uint                 num_samples;
-#ifdef CL_VERSION_2_0 && __CL_HAS_ANON_STRUCT__
+#ifdef CL_VERSION_2_0
+#if __CL_HAS_ANON_STRUCT__
 #ifdef _MSC_VER
 #if _MSC_VER >= 1500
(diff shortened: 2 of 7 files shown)


bashbaug (changes requested):
Thanks!  This is definitely an improvement.  I'm still seeing a few build failures with `/Za` with these changes though.  Note, I've found these by modifying the tests to pass `/Za`, which may be a good thing to do anyhow:

```
if (NOT MSVC)
add_compile_options(-Wall -Wextra -Werror -pedantic -Wno-format)
else ()
add_compile_options(/Za)
endif ()
```

1. There's a related code section at the end of `cl_platform.h` to pop the warning that needs a similar check for `! defined(__STDC__)`:

```c
#if defined( _WIN32) && defined(_MSC_VER) && ! defined(__STDC__)
    #if _MSC_VER >=1500
    #pragma warning( pop )
    #endif
#endif
```

2. There's a section of `cl.h` that also uses an anonymous union and is failing with `/Za`.  Note that this use is only for OpenCL 2.0 or newer.

```
#ifdef CL_VERSION_2_0
#ifdef __GNUC__
    __extension__   /* Prevents warnings about anonymous union in -pedantic builds */
#endif
#ifdef _MSC_VER
#pragma warning( push )
#pragma warning( disable : 4201 ) /* Prevents warning about nameless struct/union in /W4 /Za builds */
#endif
    union {
#endif
      cl_mem                  buffer;
#ifdef CL_VERSION_2_0
      cl_mem                  mem_object;
    };
#ifdef _MSC_VER
#pragma warning( pop )
#endif
#endif
```

Should we just pick between `buffer` and `mem_object` if we don't have anonymous structs?

MathiasMagnus:
@bashbaug I don't know about the image part of the spec enough to tell the consequences of chosing one over the other, so I'd make it up to you.

bashbaug:
Thanks!  This is a definite improvement and (1) is fixed for me.

I think we should:

* File an issue for (2) so we don't lose track of it.  This isn't quite as simple of a fix.  Gut feeling is that we'll need to pull in some of the anonymous struct / union code from `cl_platform.h` to `cl.h` and pick one of the two union members when anonymous unions aren't supported.  I don't have a strong opinion which one, since they both involve tradeoffs.
* Commit these changes as-is since they fix the pre-OpenCL 2.0 build and address (2) in a subsequent PR.

bashbaug (dismissed):
See comment: https://github.com/KhronosGroup/OpenCL-Headers/pull/116#issuecomment-700213694

I think we should merge this PR as-is and address the OpenCL 2.0 and newer anonymous union in a subsequent PR.

CL/cl_gl_ext.h
```diff
@@ -30,7 +30,7 @@ extern "C" {
 
 extern CL_API_ENTRY cl_event CL_API_CALL
 clCreateEventFromGLsyncKHR(cl_context context,
-                           cl_GLsync  cl_GLsync,
+                           cl_GLsync  /*cl_GLsync*/,
```
bashbaug, line 33: Minor: I'd prefer to use a different name for this parameter vs. commenting it out.  Maybe `sync`, which is used in the [function description](https://bashbaug.github.io/OpenCL-Docs/html/OpenCL_Ext.html#cl_khr_gl_event-new-procedures-and-functions) in the extension spec?

This comment can be addressed later if needed and shouldn't hold up this PR.

bashbaug (approved):
LGTM with one minor comment.  Thanks!

bashbaug:
I did a bit more digging into the anonymous structs / unions issue with `cl_icd.h` and discovered that this issue will occur with any test that includes the DX headers and tries to compile with `/Za`, it's just that we don't have specific tests for these headers.

I'll create an issue to track this.  As far as this PR goes, I think this means that the changes to `cl_d3d10.h`, `cl_d3d11.h`, and `cl_dx9_media_sharing.h` aren't needed and can be reverted, but they aren't hurting and could be left alone.

CL/cl_gl_ext.h
MathiasMagnus: No biggie. Changed.

Nuullll:
Hi, @MathiasMagnus 
I think this patch probably misused `__CL_HAS_ANON_STRUCT__` on **anonymous unions** in `CL.h`, and violated [OpenCL 2.0 spec](https://www.khronos.org/registry/OpenCL/specs/opencl-2.0.pdf) which explicitly allows the usage of `cl_image_desc.mem_object`

If `__CL_HAS_ANON_STRUCT__` is defined as 0 (I'll give an example with a modern compiler later), then `mem_object` member of `cl_image_desc` is apparently gone even with `CL_VERSION_2_0` defined, which breaks application codes.

`g++` will define `__STRICT_ANSI__` when standard option is set as `-std=c++XX`, according to `cl_platform.h`
```cpp
/* Define capabilities for anonymous struct members. */
#if !defined(__cplusplus) && defined(__STDC_VERSION__) && __STDC_VERSION__ >= 201112L
#define  __CL_HAS_ANON_STRUCT__ 1
#define  __CL_ANON_STRUCT__
#elif defined( __GNUC__) && ! defined( __STRICT_ANSI__ )
#define  __CL_HAS_ANON_STRUCT__ 1
#define  __CL_ANON_STRUCT__ __extension__
#elif defined( _WIN32) && defined(_MSC_VER) && ! defined(__STDC__)
    #if _MSC_VER >= 1500
   /* Microsoft Developer Studio 2008 supports anonymous structs, but
    * complains by default. */
    #define  __CL_HAS_ANON_STRUCT__ 1
    #define  __CL_ANON_STRUCT__
   /* Disable warning C4201: nonstandard extension used : nameless
    * struct/union */
    #pragma warning( push )
    #pragma warning( disable : 4201 )
    #endif
#else
#define  __CL_HAS_ANON_STRUCT__ 0
#define  __CL_ANON_STRUCT__
#endif
```

Therefore, when we compile with simple commands like `g++ -std=c++14`, the `__CL_HAS_ANON_STRUCT__` will be 0, and `cl_image_desc` will lose the field `mem_object`, breaking current application codes.

Notes: 

- C11 starts to support both anon union and anon struct
- C++ standard always supports anon union
- C++ standard doesn't support anon struct

IMO, we shouldn't guard anon union usages in `cl.h` with `__CL_HAS_ANON_STRUCT__`, since anon union and anon struct are different things anyway.

PR #118, opened 2020-09-24
add 2 nvidia-related pci-bus device query defines

4ngry:
Add 2 constants being passed to `clGetDeviceInfo` allows to query PCI bus nvidia-specific info. That devines was used in production for several years, giving the same results as nvidia-own nvml library does.

Issue #113, opened 2020-09-15
Generate Extension Headers


bashbaug:
I'm mostly raising this issue for awareness and to avoid duplicated work.

I started experimenting with generated extension headers using the cl.xml file over the weekend and I'm pretty close to having something workable.  In addition to the nice consistency benefits, this also provides a mechanism to validate that the XML file is correct for extensions, and enables easy additions for things like a symbolic name for extensions (#111) and preprocessor defines for extensions (#86).

Note that I've only done extension headers for now, and not the core API headers.  The extension headers were easier (fewer conditional rules) and are faster moving, so I think they provide a bigger bang-for-the-buck, and I don't plan to generate the core API headers at this time.

The few issues I still need to resolve are:

- [ ] What should we do for the extensions that require external headers, such as the GL, EGL, and DX extensions?  It's easy enough to omit them from the final generated headers, or we could consider a way to describe these extensions without external header dependencies.  (I know how to do this mostly - PR soon.)
- [x] The Intel extensions are in their own file with a unique copyright header.  I'll look into this.  (Fixed by #129).
- [x] I need to figure out how to add the special-case handling for `CL_DEVICE_DOUBLE_FP_CONFIG` (only for pre-OpenCL 1.2) and `cl_kernel_sub_group_info` (only for pre-OpenCL 2.1).  (See https://github.com/KhronosGroup/OpenCL-Docs/issues/463, https://github.com/KhronosGroup/OpenCL-Docs/pull/469)
- [ ] It looks like the macros for `cl_khr_extended_versioning` are not in the XML file, so they need to be added and I need to figure out how to properly generate code for them.  (I know how to do this - PR soon.)
- [x] Decide what to do with `cl_gl_ext.h` (see #145, fixed by #156).
- [ ] Ensure extension API prefix and suffixes are correct and consistent (see #153, #158).

kpet:
That's awesome! This would be a big enabler :).

Regarding the missing `cl_khr_extended_versioning` definitions, I think it's mostly because we still haven't merged the spec PR (https://github.com/KhronosGroup/OpenCL-Docs/pull/218).

bashbaug:
I think I got this mostly working over the holidays.  I found a few additional minor issues (see #145 and #146) but I'm able to generate headers that pass all of the header tests now.  Yay!

[...]

PR #119, opened 2020-09-24
Amd specific exts

4ngry:
Add AMD specific APIs and defines available for a long time under https://github.com/RadeonOpenCompute/ROCm-OpenCL-Runtime.

### DIFF
--- a/CL/cl_ext.h
+++ b/CL/cl_ext.h
@@ -230,6 +230,16 @@ typedef CL_API_ENTRY cl_command_queue
 
 
 /*********************************
+* cl_amd_device_memory_flags *
+*********************************/
+#define cl_amd_device_memory_flags 1
+#define CL_MEM_USE_PERSISTENT_MEM_AMD               (1 << 6)  // Alloc from GPU's CPU visible heap
+
+/* cl_device_info */
+#define CL_DEVICE_MAX_ATOMIC_COUNTERS_EXT           0x4032
+
+
+/*********************************
 * cl_amd_device_attribute_query *
 *********************************/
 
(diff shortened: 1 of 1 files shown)


kpet:
Thanks for contributing this! I agree it would be great to have these definitions upstream. A few comments:
- It seems this PR has the NVidia definitions from #118 as well. I assume that was not intended.
- You will need to sign the CLA before we can accept the contribution.
- These being vendor definitions, a review from AMD and NVidia will be needed (cc @nikhiljnv @b-sumner)
- These definitions should be added to the XML file describing all API definitions (https://github.com/KhronosGroup/OpenCL-Docs/blob/master/xml/cl.xml)

nikhiljnv:
l looked at NVIDIA extension specific defines. Seem ok to me. But, there are few more changes that I need to upstream. So, wouldn't mind dropping these from this change and I can take care of this.

4ngry:
Ok, it seems that I signed CLA already but some misconfiguration with git username and github username prevents the signing.
I will do the required work on xml documentation too.

bashbaug:
I don't have much to add to the comments above, so I'll just reiterate:

* Thanks for doing this!  It'll be great to have all of these extensions in one place vs. scattered in vendor SDKs.
* Since these are AMD extensions they should have a review from someone from AMD (re-ping @b-sumner @saleelk).
* We also need to add these definitions to the API XML file, especially if we move to auto-generated extension headers (see #113).  I can take care of this, just re-mentioning it for completeness.
* There are a few additional extensions mentioned in [this gist](https://gist.github.com/ranisalt/af4f803f498ab8b3190d3754b2c78ff6#file-rocm-opencl-headers-patch) that we could consider adding to this PR, or could add in a subsequent PR.

Thanks again!

Issue #121, opened 2020-10-03
[OpenCL 2.0][cl.h with C++] struct `cl_image_desc` is missing field `mem_object`


Nuullll:
This is caused by PR #116 

Minimal reproducer:
```cpp
// test.cpp
#include <CL/cl.h>

int main() {
  cl_image_desc image_desc;
  (void)image_desc.mem_object;
  return 0;
}
```

Then compile with
```bash
g++ -I$(pwd) -std=c++14 test.cpp
```

Error:
```text
note: #pragma message: cl_version.h: CL_TARGET_OPENCL_VERSION is not defined. Defaulting to 220 (OpenCL 2.2)
test.cpp: In function ‘int main()’:
test.cpp:6:20: error: ‘cl_image_desc’ {aka ‘struct _cl_image_desc’} has no member named ‘mem_object’
   (void)image_desc.mem_object;
                    ^~~~~~~~~~
```

Nuullll:
Reverting the change on `cl.h` could fix this issue.
Explained here: https://github.com/KhronosGroup/OpenCL-Headers/pull/116#issuecomment-703090867

bashbaug:
I don't think we should completely revert the change to `cl.h` because there are some situations where anonymous structs and unions are not supported.

I do wonder though if we should be trying to identify these cases where anonymous structs and unions ARE NOT supported to skip them vs. trying to identify the cases where anonymous structs and unions ARE supported to enable them.  In other words, flip the conditions so anonymous structs and unions are supported by default, since this is likely to be a more common case (and an easier fix for the compilers that do not support anonymous structs and unions).

bashbaug:
Note, I think we also need to differentiate between anonymous structs and anonymous unions:

https://stackoverflow.com/questions/12355118/which-standards-allow-anonymous-structs-and-unions-in-c-and-c/12355215

Anonymous unions are supported more broadly than anonymous structs.

Nuullll:
Thanks! @bashbaug
Differentiating anonymous structs/unions and flipping the condition sounds to be a decent solution.

I'm also wondering what should the expected behavior be when anonymous structs/unions are not supported. Say we are compiling OpenCL 2.0 application codes using MS C compiler with the `/Za` option, where MS extensions to ANSI C89/ISO C90 are disabled so that anonymous unions are not supported. Then:
[...]

PR #122, opened 2020-10-03
Revert using of __CL_HAS_ANON_STRUCT__ in cl.h

Nuullll:
This PR addresses #121, by reverting changes on `cl.h` of PR #116

### DIFF
--- a/CL/cl.h
+++ b/CL/cl.h
@@ -134,27 +134,21 @@ typedef struct _cl_image_desc {
     cl_uint                 num_mip_levels;
     cl_uint                 num_samples;
 #ifdef CL_VERSION_2_0
-#if __CL_HAS_ANON_STRUCT__
+#ifdef __GNUC__
+    __extension__   /* Prevents warnings about anonymous union in -pedantic builds */
+#endif
 #ifdef _MSC_VER
-#if _MSC_VER >= 1500
 #pragma warning( push )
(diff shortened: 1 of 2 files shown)


Nuullll:
I also noticed that the compile options for MSVC and gcc/clang are a little different in test builds:
```cmake
    # ICD on Windows uses system headers, which aren't strictly ANSI conformant
    # and trigger warnings 
    if(${NAME} STREQUAL cl_icd_h)
      set(MSVC_COMPILE_DEFS /W4 /WX)
    else()
      set(MSVC_COMPILE_DEFS /W4 /WX /Za)
    endif()
    target_compile_options(${TEST_EXE}
      PUBLIC
        $<$<NOT:$<CXX_COMPILER_ID:MSVC>>:-Wall -Wextra -Werror -pedantic -Wno-format>
        $<$<CXX_COMPILER_ID:MSVC>:${MSVC_COMPILE_DEFS}>
    )
```

gcc uses `-std=gnu90` before v5.1 and `-std=gnu11` starting from v5.1 (released in 2015). `gnuXX` means `cXX` standard with GNU extensions. And C11 starts to support anon unions.

Setting `/Za` will force MSVC conform with the ancient C89 standard, which doesn't allow anon unions. That's why the test build was failing.

I guess the anon union used in `cl_image_desc` is for OpenCL 1.2 compability, which is probably inevitable. So why not enable MSVC's extensions to offer same capability as GCC for anon unions?

Nuullll:
Abandoned. See PR #125.

PR #123, opened 2020-10-05
Add testing for cl_d3d10.h, cl_d3d11.h, and cl_dx9_media_sharing.h

bashbaug:
Adds missing testing for cl_d3d10.h, cl_d3d11.h, and cl_dx9_media_sharing.h.

PR #125, opened 2020-10-06
re-enable anonymous unions by default

bashbaug:
This is a slight relaxation of the changes in #116 to re-enable anonymous unions by default (but not anonymous structs), and to disable anonymous unions only when they are known to not be supported, which is currently only for MSVC builds with `/Za`.

See also #121 

It's probably easiest to review these changes by [comparing against the previous commit](https://github.com/KhronosGroup/OpenCL-Headers/compare/9d17180a33566500551ff98e9eba62b0867a6f64...bashbaug:relax-anonymous-union) before #116.  The main difference is the addition of an `#ifdef` block to bypass the anonymous union for `/Za' builds.

This change passed all of my testing on Windows, including the tests I added in #123.

### DIFF
--- a/CL/cl.h
+++ b/CL/cl.h
@@ -134,27 +134,29 @@ typedef struct _cl_image_desc {
     cl_uint                 num_mip_levels;
     cl_uint                 num_samples;
 #ifdef CL_VERSION_2_0
-#if __CL_HAS_ANON_STRUCT__
-#ifdef _MSC_VER
-#if _MSC_VER >= 1500
+#if defined(__GNUC__)
+    __extension__                   /* Prevents warnings about anonymous union in -pedantic builds */
+#endif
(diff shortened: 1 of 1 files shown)


Nuullll:
Thanks for your effort!

This is working for me on linux. And I'll test on windows later.

I think we should also `#undef` anon-struct macros at the end of `cl_platform.h`, since they are not used elsewhere:
https://github.com/KhronosGroup/OpenCL-Headers/pull/122/files#diff-9771949e68ce01ebe73dfdb5d75e8ab8

MathiasMagnus (approved):
LGTM

bashbaug:
Thanks!  I'm going to merge these changes now.  We can pick up the `#undef` changes in a separate PR.

PR #128, opened 2020-10-13
switch the default version for the OpenCL headers to OpenCL 3.0

bashbaug:
Since OpenCL 3.0 is now final, we should switch the default OpenCL version for the headers if no OpenCL version is provided to OpenCL 3.0 instead of OpenCL 2.2.


alycm (approved):
LGTM.

PR #129, opened 2020-10-16
move Intel extensions into common files

bashbaug:
This PR moves the enum defines, typedefs, and function prototypes for Intel extensions from Intel-specific files into common extension files.  This will simplify auto-generation of extension headers (see #113).

No files are deleted as part of this PR.  Instead, the Intel extension files now include the common extension files and emit a warning that the common extension file should be directly included instead.  We can consider removing the Intel files later, if desired, after projects have had an opportunity to update their include files.

Two notes worth calling out specifically:

* I haven't modified the VA_API sharing extension files since there is no common file to move this extension to.
* The Intel copyright header and disclaimers were purposefully removed as they are not necessary for these extension headers.


EwanC (dismissed):
Out of curiosity was there a reason the Intel extensions were in a separate files to being with? Looks like it's the only vendor to do so

bashbaug:
> Out of curiosity was there a reason the Intel extensions were in a separate files to being with?

I tried tracking this down but I'm afraid the reasons are lost to time.  I suspect it's because we originally had a separate file of extensions that was added unchanged, but that's just a guess.

I suppose we could consider breaking the extension headers out by vendor consistently, but I'm not sure what this would buy us and I don't think any of the other Khronos APIs follow this pattern.

I'll wait another couple of days before merging just in case someone else wants to give it a review.  Thanks!

CL/cl_dx9_media_sharing_intel.h
```diff
@@ -13,158 +13,6 @@
...
-    cl_event*                   event) CL_EXT_SUFFIX__VERSION_1_1;
-
-#ifdef __cplusplus
-}
-#endif
-
-#endif  /* __OPENCL_CL_DX9_MEDIA_SHARING_INTEL_H */
 
+#include <CL/cl_dx9_media_sharing.h>
+#pragma message("The Intel DX9 media sharing extensions have been moved into cl_dx9_media_sharing.h.  Please include cl_dx9_media_sharing.h directly.")
```
kpet, line 18: Suggest using `#warning` instead.

CL/cl_ext_intel.h
```diff
@@ -14,718 +14,6 @@
...
-/* cl_mem_flags */
-#define CL_MEM_FORCE_HOST_MEMORY_INTEL                      (1 << 20)
-
-#ifdef __cplusplus
-}
-#endif
-
-#endif /* __CL_EXT_INTEL_H */
+#include <CL/cl_ext.h>
+#pragma message("The Intel extensions have been moved into cl_ext.h.  Please include cl_ext.h directly.")
```
kpet, line 19: Same here

tests/CMakeLists.txt
```diff
@@ -35,7 +35,6 @@ endfunction(add_header_test)
 add_header_test(cl_h test_cl.h.c)
 add_header_test(cl_egl_h test_cl_egl.h.c)
 add_header_test(cl_ext_h test_cl_ext.h.c)
-add_header_test(cl_ext_intel_h test_cl_ext_intel.h.c)
```
kpet, line 38: We should keep testing the file as long as it exists.

kpet (changes requested):
Looks fine overall but at least the test comment needs to be addressed.

CL/cl_dx9_media_sharing_intel.h
bashbaug: Thank you for the suggestion, but I'd prefer to use `#pragma message` instead, to avoid breaking builds with warnings-as-errors enabled.  This is also consistent with the warning-like messages in `cl_version.h` when the header version is not set or is set incorrectly.

(As an added bonus, this is also required for testing, see below.)

CL/cl_ext_intel.h
bashbaug: See above.

tests/CMakeLists.txt
bashbaug: I've added back testing for this file.  Note, this means that the message about extensions being moved is emitted when tests are built, as expected.  I just want to confirm this behavior is OK:

```sh
[ 20%] Building C object CMakeFiles/cl_ext_intel_h_300.dir/test_cl_ext_intel.h.c.o
In file included from /home/bashbaug/git/OpenCL-Headers/tests/test_cl_ext_intel.h.c:19:0:
/home/bashbaug/git/OpenCL-Headers/tests/../CL/cl_ext_intel.h:19:9: note: #pragma message: The Intel extensions have been moved into cl_ext.h.  Please include cl_ext.h directly.
 #pragma message("The Intel extensions have been moved into cl_ext.h.  Please include cl_ext.h directly.")
         ^~~~~~~
[ 21%] Linking C executable cl_ext_intel_h_300
```

CL/cl_ext.h
```diff
@@ -899,6 +899,682 @@ clSetKernelExecInfoARM(cl_kernel            kernel,
...
+#define CL_AVC_ME_SLICE_TYPE_INTRA_INTEL                    0x2
+
+#define CL_AVC_ME_INTERLACED_SCAN_TOP_FIELD_INTEL           0x0
+#define CL_AVC_ME_INTERLACED_SCAN_BOTTOM_FIELD_INTEL        0x1
+
+/*******************************************
+* cl_intel_unified_shared_memory extension *
+********************************************/
+
+/* These APIs are in sync with Revision Q of the cl_intel_unified_shared_memory spec! */
```
EwanC, line 1263: nitpick: should this be bumped to revision R?

EwanC (approved):

CL/cl_ext.h
bashbaug: It's been a while since I've created this PR!  In all seriousness, rev R fixed a typo in a table, so it is probably true that these headers are in sync with rev R rather than rev Q, but the differences between the two versions are very minor.

bashbaug:
This PR has been open for a while.  @kpet, do the latest changes look OK?  If so, can this be merged please?

CL/cl_dx9_media_sharing_intel.h
kpet: Yes, and in the meantime, https://github.com/KhronosGroup/OpenCL-CLHPP/pull/122 happened :).

tests/CMakeLists.txt
kpet: Yes, that's fine.

kpet (approved):
All looks fine. Merging.

PR #134, opened 2020-11-10
Added missing define for cl_img_use_gralloc_ptr.

Jeremy-Kemp:
Accompanies KhronosGroup/OpenCL-Docs#487.


bashbaug (approved):
Matches the spec and XML file - LGTM!

PR #135, opened 2020-11-10
Added definitions for cl_img_generate_mipmap.

Jeremy-Kemp:
Accompanies KhronosGroup/OpenCL-Docs#489


bashbaug (approved):
LGTM and this matches the spec, however `clEnqueueGenerateMipmapIMG` does not appear to be in the XML file.  We should add it at some point.

Jeremy-Kemp:
Thanks for spotting that @bashbaug. I'll submit another PR shortly.

PR #136, opened 2020-11-11
Added definitions for cl_img_mem_properties.

Jeremy-Kemp:
Accompanies KhronosGroup/OpenCL-Docs#491


bashbaug (approved):
LGTM!

bashbaug:
It looks like the Travis job is failing since Khronos is out of credits.

I have verified that all tests pass locally so I am going to merge this change as-is to unblock progress.  We really need to get these CI issues resolved though...

PR #138, opened 2020-11-20
enable testing via GitHub actions

bashbaug:
This sets up an alternative CI workflow using GitHub actions.  It is intentionally very similar to the existing Travis workflow:

* It tests Windows, Linux, and OSX.
* It builds all tests with Debug compile options (it's easy to change this if we want to).
* It makes sure all tests run correctly.

The main difference is that this only tests the default compiler for each OS, meaning gcc and Clang are not tested separately.  We can add this, I think, but it'll take a bit more work.

I haven't disabled any of the Travis CI workflows in this PR, but if GitHub actions works just as well (or better) we can look into disabling it in the future.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -9,7 +9,7 @@ jobs:
   build:
     strategy:
       matrix:
-        os: [ubuntu-latest, windows-latest]
+        os: [ubuntu-latest, windows-latest, macos-latest]
 
     runs-on: ${{matrix.os}}
 
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -1,6 +1,16 @@
 name: CMake
 
-on: [push]
+on:
+  push:
+    branches:
+      - master
+    paths-ignore:
+      - '**.md'
+  pull_request:
+    branches:
+      - master
+    paths-ignore:
+      - '**.md'
 
 env:
   BUILD_TYPE: Debug
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -1,0 +1,35 @@
+name: CMake
+
+on: [push]
+
+env:
+  BUILD_TYPE: Debug
+
+jobs:
+  build:
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-latest]
+
+    runs-on: ${{matrix.os}}
+
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Create Build Directory
+      run: cmake -E make_directory ${{runner.workspace}}/build
+
+    - name: Run CMake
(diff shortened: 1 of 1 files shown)


.github/workflows/cmake.yml
```diff
@@ -0,0 +1,45 @@
...
+jobs:
+  build:
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-latest, macos-latest]
+
+    runs-on: ${{matrix.os}}
+
+    steps:
+    - uses: actions/checkout@v2
```
EwanC, line 27: not quite understanding the syntax here even with the help of the [docs](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#example-using-versioned-actions).

`actions/checkout@v2`  I think this is checking out the branch of the repo from the push/pull request, but why version `2`?

oddhack: I just cut & pasted from the GH Actions docs when I started playing with it, but I assume @v2 is the current version of the builtin action and somehow incompatible with the previous version.

.github/workflows/cmake.yml
```diff
@@ -0,0 +1,45 @@
+name: CMake
+
+on:
+  push:
+    branches:
+      - master
```
oddhack, line 6: Wouldn't you want to run on non-default branches when pushed to the repo, as well? ISTM this is never going to catch problems before they're merged to master.

.github/workflows/cmake.yml
MathiasMagnus: Indeed, `uses` clauses can use complex (or simple) actions provided by others. Instead of manually providing `git clone` as a shell command, one can import that in a slightly more robust way with customization points exposed as parameters to `uses`. Introduction [here](https://docs.github.com/en/free-pro-team@latest/actions/learn-github-actions/introduction-to-github-actions#understanding-the-workflow-file) and the specific action and it's parameters [here](https://github.com/marketplace/actions/checkout)

MathiasMagnus:
#107 already went through with the first run already shown under [actions](https://github.com/KhronosGroup/OpenCL-Headers/actions) in detail [here](https://github.com/KhronosGroup/OpenCL-Headers/actions/runs/381660421). 100 test cases run in 13 minutes. If that's too much, we can start cracking at reducing it, but overall GA is responsive.

I say we close this PR and open issues for GA enacted by 107. (That is if CI isn't fabulous enough already ;))

.github/workflows/cmake.yml
EwanC: I see, thanks for the links @MathiasMagnus :+1:

bashbaug:
Yup, this isn't needed anymore after #107.  Closing!

PR #140, opened 2020-12-05
fix printf format string warnings

### LINKED ISSUE #132: Resolve -Wformat warnings
Currently to pass builds cleanly, warning-free we have to disable format-related warnings. We should consider fixing those.

bashbaug:
Fixes #132 

This PR fixes the printf format string warnings by using the `PRId64` and `PRIu64` format constants.  This is admittedly a little ugly but it seems to be the most portable way of printing a 64-bit integer, and it built and ran with no warnings or failures for me on Windows and Linux.

https://en.cppreference.com/w/cpp/header/cinttypes

Since the format string warnings are fixed I've also removed `-Wno-format` from the CI scripts.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -32,7 +32,7 @@ jobs:
         -G "${{matrix.GEN}}"
         -D BUILD_TESTING=ON
         -D CMAKE_BUILD_TYPE=${{matrix.VAR}}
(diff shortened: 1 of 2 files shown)


MathiasMagnus (approved):
The link to the printf macros is slightly off (C version [here](https://en.cppreference.com/w/c/types/integer)) but the changes look good and removing warnings is always welcome. Clean builds for everyone!

alycm (approved):
LGTM.

jrprice (approved):

Issue #124, opened 2020-10-06
Stop using reserved C/C++ reserved identifiers


kpet:
The headers are using a number of identifiers that are reserved in C/C++ (e.g. those beginning with `__`, see https://en.cppreference.com/w/c/language/identifier). We should rename all of these.

PR #148, opened 2021-01-26
Move cl_icd_layer.h from OpenCL-ICD-Loader to OpenCL-Headers.

Kerilk:
As discussed in https://github.com/KhronosGroup/OpenCL-Docs/pull/521, moving the layer definition header here.


CL/cl_icd_layer.h
```diff
@@ -0,0 +1,61 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * OpenCL is a trademark of Apple Inc. used under license by Khronos.
+ */
+
+#ifndef __ICD_LAYER_H
```
bashbaug, line 19: This is fairly minor, but since we have no backward compatibility concerns (yet), I think we should use an include guard that is not technically a reserved identifier (see https://github.com/KhronosGroup/OpenCL-Headers/issues/124).  Could we use `OPENCL_CL_ICD_LAYER_H` instead?  This seems to be the informal convention adopted by `cl_half.h` and `cl_icd.h`.

bashbaug (approved):
LGTM.  Just one comment that could be addressed later (but would be nice to address now).

CL/cl_icd_layer.h
Kerilk: That's a very good suggestion Ben, thanks for catching this.

bashbaug (dismissed):

alycm (dismissed):
LGTM.

I did wonder if the header should be `cl_loader_layer.h` or similar to match the `cl_loader_layers` extension name.   However, there is already a `cl_icd.h` header, so following that makes sense.

Kerilk:
> LGTM.
> 
> I did wonder if the header should be `cl_loader_layer.h` or similar to match the `cl_loader_layers` extension name. However, there is already a `cl_icd.h` header, so following that makes sense.

Thanks for raising this point, the file was renamed to `cl_layer.h`.

bashbaug (approved):
LGTM - thanks!

PR #152, opened 2021-02-09
Add definitions for cl_arm_controlled_kernel_termination

kpet:

### DIFF
--- a/CL/cl_ext.h
+++ b/CL/cl_ext.h
@@ -922,12 +922,12 @@ clSetKernelExecInfoARM(cl_kernel            kernel,
 #define CL_EVENT_COMMAND_TERMINATION_REASON_ARM 0x41ED
 
 /* Values returned for event termination reason query */
-typedef enum {
-    CL_COMMAND_TERMINATION_COMPLETION_ARM = 0,
-    CL_COMMAND_TERMINATION_CONTROLLED_SUCCESS_ARM = 1,
-    CL_COMMAND_TERMINATION_CONTROLLED_FAILURE_ARM = 2,
-    CL_COMMAND_TERMINATION_ERROR_ARM = 3
-} cl_command_termination_reason_arm;
+typedef cl_uint cl_command_termination_reason_arm;
+
+#define CL_COMMAND_TERMINATION_COMPLETION_ARM  0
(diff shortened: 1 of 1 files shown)


CL/cl_ext.h
```diff
@@ -901,6 +901,34 @@ clSetKernelExecInfoARM(cl_kernel            kernel,
...
+/* cl_event_info */
+#define CL_EVENT_COMMAND_TERMINATION_REASON_ARM 0x41ED
+
+/* Values returned for event termination reason query */
+typedef enum {
+    CL_COMMAND_TERMINATION_COMPLETION_ARM = 0,
+    CL_COMMAND_TERMINATION_CONTROLLED_SUCCESS_ARM = 1,
+    CL_COMMAND_TERMINATION_CONTROLLED_FAILURE_ARM = 2,
+    CL_COMMAND_TERMINATION_ERROR_ARM = 3
+} cl_command_termination_reason_arm;
```
bashbaug, line 930: This isn't broken, but we haven't used this pattern in the headers before.

Could we instead typedef `cl_command_termination_reason_arm` as a `cl_uint` (?), and then define these values instead?

This would be more consistent with the other extensions and would fit better with generated headers (see #113).  It _might_ also avoid possible problems with the size of an enum.

bashbaug (commented):
Generally looks good to me and matches the spec in https://github.com/KhronosGroup/OpenCL-Docs/pull/557, but I do have one question.  Thanks!

CL/cl_ext.h
kpet: Yes, we could. There are advantages to using an enum (better diagnostics, clarity, etc) and disadvantages as you've pointed out. If moving from the enum makes generation easier then I'm happy to do that.

bashbaug (dismissed):
Thanks, the enum part LGTM now.

Is this bitfield type still needed?  It's in the XML PR:

```c
typedef cl_bitfield cl_device_controlled_termination_capabilities_arm;
```

If not, feel free to merge anytime.

kpet:
Yes, it is! There was back and forth on this and it didn't make it into the PR. Thanks for catching that!

bashbaug (approved):

Issue #153, opened 2021-02-09
Proposed tidy up of CL_{API,EXT}_{PREFIX,SUFFIX}* definitions


kpet:
These definitions are used to prefix/suffix function declarations, for example to handle deprecation warnings. They currently fall into two categories:
- Per-version definitions (`CL_{API,EXT}_SUFFIX__{EXPERIMENTAL,VERSION*}`). These are always defined to an empty string.
- Per-version deprecation definitions (`CL_EXT_SUFFIX__VERSION_*_DEPRECATED`). These are defined to compiler-specific values that result in warnings being produced when deprecated functions are used. Note that there are no `CL_API_SUFFIX__VERSION_*_DEPRECATED` definitions.

### Problems

- `CL_EXT` and `CL_API` prefixes are used inconsistently and core API function declarations are marked as deprecated with `CL_EXT` definitions
- The distinction doesn't seem to serve any purpose and is confusing

As another data point, extension functions use `CL_API_{CALL,ENTRY}` in their declarations.

### Proposed resolution

- Use `CL_API_{PREFIX,SUFFIX}` for core declarations as well as extensions and remove all `CL_EXT_{PREFIX,SUFFIX}` definitions.

OpenCL-Docs and OpenCL-CLHPP would need changing as well.

bashbaug:
Switching to the `CL_API_{PREFIX,SUFFIX}` vs. `CL_EXT_{PREFIX,SUFFIX}` makes sense to me, +1.

Is there a usage model for the `CL_API_SUFFIX__{EXPERIMENTAL,VERSION*}` definitions or can they be removed?

kpet:
Ok, I'll do that.

>Is there a usage model for the CL_API_SUFFIX__{EXPERIMENTAL,VERSION*} definitions or can they be removed?

I actually started using them in https://github.com/KhronosGroup/OpenCL-Headers/pull/154 (which I'd like to get in before the tidy up as they would conflict).

bashbaug:
I was actually looking at #154 which prompted my question.  Right now it looks like we have a possible use for prefix and suffix defiitions generally (for things like #154), and for prefix and suffix defiitions for deprecations for specific versions, but I don't currently see a need for prefix and suffix definitions for specific versions that are not tied to deprecations.

If so, would be fine with:

* General prefix and suffix definitions that are not version-specific and are used for most APIs.  These would replace the current version-specific prefix and suffix definitions for non-deprecated APIs.
[...]

PR #154, opened 2021-02-09
Add support for user-supplied prefix/suffix in function declarations

kpet:
This change adds the ability for a user of the headers to pass
prefixes or suffixes to be used in the function prototypes via
(respectively) CL_API_PREFIX_USER and CL_API_SUFFIX_USER.

This can be used to pass additional attributes (e.g. to use
weak symbols).

Signed-off-by: Kévin Petit <kpet@free.fr>


kpet:
Also see the (related) proposed tidy up on https://github.com/KhronosGroup/OpenCL-Headers/issues/153

CL/cl_platform.h
```diff
@@ -41,22 +41,33 @@ extern "C" {
...
+#ifndef CL_API_SUFFIX_USER
+#define CL_API_SUFFIX_USER
+#endif
+
+#ifndef CL_API_PREFIX_USER
+#define CL_API_PREFIX_USER
+#endif
+
+#define CL_API_SUFFIX_COMMON CL_API_SUFFIX_USER
+#define CL_API_PREFIX_COMMON CL_API_PREFIX_USER
```
bashbaug, line 53: I don't know how much we care about this, but I think we could avoid defining these symbols and just use the user-defined prefix and suffixes directly.  Alternatively, we could `#undef` these symbols after they're used.

As a separate issue (not for this PR), I think we should also formally reserve all of the `CL_API_XXX` names in the spec for usage in the headers.

bashbaug (approved):
LGTM, just one small comment / question that doesn't need to hold up merging.

CL/cl_platform.h
kpet: I added these to make it easier to manage common definitions in the future. I agree they're not strictly required for this change. We could indeed `#undef` them after but I'm not sure it matters too much, especially if we're reserving all the `CL_API*` names (I've added this to a checlist in #153).

kpet:
I'll merge this one as is to get forward progress but I'm happy to discuss other changes as part of #153. I've broadened the scope of that issue to cover all related problems and added a checklist there so we don't lose track of what needs doing.

PR #158, opened 2021-02-16
Replace uses of CL_EXT_{PRE,SUF}FIX* with CL_API_{PRE,SUF}FIX*

kpet:
Contributes to #153

Signed-off-by: Kévin Petit <kevin.petit@arm.com>


CL/cl_platform.h
```diff
@@ -53,80 +53,80 @@ extern "C" {
 #define CL_API_PREFIX_COMMON CL_API_PREFIX_USER
 
 #define CL_API_SUFFIX__VERSION_1_0 CL_API_SUFFIX_COMMON
-#define CL_EXT_SUFFIX__VERSION_1_0 CL_API_SUFFIX_COMMON
+#define CL_API_SUFFIX__VERSION_1_0 CL_API_SUFFIX_COMMON
```
jrprice, line 56: Looks like a duplicate definition here and for the other versions below.

jrprice (changes requested):

CL/cl_platform.h
kpet: Good catch. It was a bit more than a sed job :).

kpet: Now fixed.

jrprice (approved):

PR #162, opened 2021-04-07
Fixes for usage of macro CL_API_ENTRY

johnplate:
The macro `CL_API_ENTRY` is obviously intended for storage-class attributes, e.g. `__declspec(dllexport)`. Those attributes are not allowed in `typedef` declarations and will result in compilation errors. I have also submitted a [PR](https://github.com/KhronosGroup/OpenCL-Docs/pull/591) to fix the documentation.

Further `CL/cl_platform.h` contains changes to allow the macros `CL_API_ENTRY`, `CL_API_CALL` and `CL_CALLBACK` to be overridden externally.

Additionally `CL/cl_ext.h` contains fixes for the function declarations of extensions `cl_APPLE_SetMemObjectDestructor` and `cl_APPLE_ContextLoggingFunctions`.

### DIFF
--- a/CL/cl_d3d10.h
+++ b/CL/cl_d3d10.h
@@ -75,7 +75,7 @@ typedef cl_uint cl_d3d10_device_set_khr;
 
 /******************************************************************************/
 
-typedef CL_API_ENTRY cl_int (CL_API_CALL *clGetDeviceIDsFromD3D10KHR_fn)(
+typedef cl_int (CL_API_CALL *clGetDeviceIDsFromD3D10KHR_fn)(
     cl_platform_id             platform,
     cl_d3d10_device_source_khr d3d_device_source,
     void *                     d3d_object,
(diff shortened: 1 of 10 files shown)


johnplate:
The linux CI builds failed because gcc-7 is missing.

bashbaug (approved):
That's a little embarassing.  Changes look good to me.

I've also updated the generated headers in #161 so they don't include `CL_API_ENTRY` on the typedef.

@jrprice, do you mind taking a quick look before merging?

@MathiasMagnus, do we need to explicitly install gcc 7 on Linux now?  If this isn't an easy fix we should disable this config for now since the gcc 7 failure stops other testing on Linux.

MathiasMagnus:
> @MathiasMagnus, do we need to explicitly install gcc 7 on Linux now?  If this isn't an easy fix we should disable this config for now since the gcc 7 failure stops other testing on Linux.

Seems like it. The `ubuntu-latest` image of GA has moved on to Ubuntu 20.04, which has GCC 9-10, as opposed to previous GCC 7-8-9. We should either revert from Ubuntu-latest to ubuntu-18.04 image or test using GCC 9-10, or add the `apt install gcc-8` command to the setup parts.

jrprice (approved):

jrprice:
Changes look good.

We should probably pick a specific Ubuntu version to keep things more stable. Testing with both older compilers and newer compilers sounds like a good idea, so maybe 18.04 is the best choice right now.

bashbaug:
Merging... testing will fail until we fix or disable the CI configs (see #163).

PR #163, opened 2021-04-12
disable failing CI configs

bashbaug:
We need to update our CI configs to work with recent GitHub actions changes.

This PR disables the failing configs until we do this.  Specific changes in this PR are:

1. Use the Ubuntu 18.04 image instead of the Ubuntu 20.04 (ubuntu-latest) image.
2. Disable the gcc-8 config since it is no longer included in the Ubuntu 18.04 image.
2. Disable the MacOS Ninja config since the current method of installing it is no longer working.

If this PR is merged I'll create an issue to track re-enabling whichever configs we want to re-enable.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -7,7 +7,8 @@ jobs:
     runs-on: ubuntu-18.04
     strategy:
       matrix:
-        VER: [7, 8, 9]
+        #VER: [7, 8, 9]
+        VER: [7, 9] # gcc-8 is not installed on the Ubuntu 18.04 image
         EXT: [ON, OFF]
         GEN: [Unix Makefiles, Ninja]
         VAR: [Debug, Release]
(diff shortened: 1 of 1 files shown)


MathiasMagnus (approved):
- The Ubuntu base image version shouldn't matter much (beside available compilers), since we don't use any system components. (We're building them right now.)
- The available compiler versions changing is something we have to live with, as the free tier of GA doesn't allow using custom Docker images. Installing GCC8 from `apt` is possible, but it will add a lot of CI time.
  - That would still be tolerable if we reduced the number of CI matrix entries in favor of multiple tests in one job.
- I'm not a Mac user, my only colleague who uses it is working from remote now, so all MacOS fixes are extremely time consuming now. I'd vote for disabling for now and re-add later.

oddhack:
> * The available compiler versions changing is something we have to live with, as the free tier of GA doesn't allow using custom Docker images. Installing GCC8 from `apt` is possible, but it will add a lot of CI time.

Is OpenCL-Headers using free GA and Vulkan-Docs is not? We're using our own Docker image there without problems: https://github.com/KhronosGroup/Vulkan-Docs/blob/main/.github/workflows/CI.yml

bashbaug:
I created #165 to track re-enabling the disabled CI configs, if desired.

Merging this PR for now, to unblock CI.

Issue #127, opened 2020-10-06
Which compilers, compiler options, and C/C++ standard versions do the headers support?


bashbaug:
We haven't documented which compilers and compiler options including C and C++ standard versions the headers support, which makes it difficult to determine if issues like #115 and #121 are header issues or unsupported configurations.  We should document (and test for!) the compilers and compiler options we're intending to support and be clear that other configurations may not work or have limitations.

FWIW, there is a similar issue already for the C++ bindings:
https://github.com/KhronosGroup/OpenCL-CLHPP/issues/78

MathiasMagnus:
I'm currently cooking up Github Actions CI, which I intend on propagating through the OpenCL repos (Headers, CLHPP, ICD, SDK). In the specific case of the C headers, I'm initially looking to implement this matrix:

### Linux
- COMP: GCC-7, GCC-8, GCC-9, LLVM-9, LLVM-10, LLVM-11
- STD: c90, c99, c11, c18, gnu90, gnu99, gnu11, gnu18
- GEN: GNU Makefiles, Ninja
- VAR: Debug, Release

### Windows
- COMP: MSVC-v140, MSVC-v141, MSVC-v142, LLVM-9, LLVM-10, LLVM-11
- STD: c90, c11, ms90
- GEN: MSBuild, Ninja
- VAR: Debug, Release

### MacOS
- COMP: ? (Need little help with defining a reasonable compiler set)
- STD: c90, c11, c18
- GEN: Ninja
- VAR: Debug, Release

oddhack:
Requiring support for 30 year old C standards seems excessive, but whatever you do, suggest you state it explicitly in the API spec, as Vulkan does (if there is any such statement, I couldn't easily find it).

MathiasMagnus:
@oddhack It's not for me to say. C not having a standard release for 21 years between 90 and 11 is something we have to deal with. 10 years old code _had_ to be developed against a 30 year old standard. As much as I love upgrading stuff, before calling it quits on code that's not even 10 years old... That's almost as if your compiler would tell you C++11 is so passé, I'll reject compiling it tomorrow.

oddhack:
> @oddhack It's not for me to say. C not having a standard release for 21 years between 90 and 11 is something we have to deal with.

We baselined Vulkan on C99, which is AFAIK a standard (ISO/IEC 9899:1999).

MathiasMagnus:
[...]

PR #164, opened 2021-04-14
Add missing 'stdint.h' include to 'CL/cl.h'

johnplate:
The header `CL/cl.h` is using `intptr_t`, but doesn't include the required `stdint.h` header.

On non-Windows platforms the header is already included by `CL/cl_platform.h`, but on Windows it causes a compilation error if `stdint.h` is not included before using `CL/cl.h`.

### DIFF
--- a/CL/cl.h
+++ b/CL/cl.h
@@ -20,6 +20,8 @@
 #include <CL/cl_version.h>
 #include <CL/cl_platform.h>
 
+#include <stdint.h>
+
 #ifdef __cplusplus
 extern "C" {
 #endif
--- a/CL/cl.h
+++ b/CL/cl.h
@@ -20,8 +20,6 @@
 #include <CL/cl_version.h>
 #include <CL/cl_platform.h>
 
-#include <stdint.h>
-
 #ifdef __cplusplus
 extern "C" {
 #endif
--- a/CL/cl_platform.h
+++ b/CL/cl_platform.h
@@ -19,6 +19,9 @@
 
 #include <CL/cl_version.h>
 
+/* Always required to support intptr_t in cl.h */
+#include <stdint.h>
+
 #ifdef __cplusplus
 extern "C" {
 #endif
@@ -234,8 +237,6 @@ typedef double                  cl_double;
 #define CL_INFINITY         CL_HUGE_VALF
 
 #else
-
-#include <stdint.h>
 
 /* scalar types  */
 typedef int8_t          cl_char;
--- a/CL/cl_platform.h
+++ b/CL/cl_platform.h
@@ -19,9 +19,6 @@
 
 #include <CL/cl_version.h>
 
-/* Always required to support intptr_t in cl.h */
-#include <stdint.h>
-
 #ifdef __cplusplus
 extern "C" {
 #endif
@@ -138,6 +135,12 @@ extern "C" {
 
 #if (defined (_WIN32) && defined(_MSC_VER))
 
+/* intptr_t is used in cl.h and provided by stddef.h in Visual C++, but not in clang */
+/* stdint.h was missing before Visual Studio 2010, include it for later versions and for clang */
+#if defined(__clang__) || _MSC_VER >= 1600
+    #include <stdint.h>
+#endif
+
 /* scalar types  */
 typedef signed   __int8         cl_char;
 typedef unsigned __int8         cl_uchar;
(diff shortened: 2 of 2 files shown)


bashbaug:
Hi @johnplate , thank you for continuing to submit PRs to fix issues like these!

Regarding this change:

1. I'd like to improve our testing to be sure we catch similar issues like it in the future.  I tried to do this here, by including the OpenCL headers before any other headers in our test files...

    https://github.com/bashbaug/OpenCL-Headers/tree/include-cl-header-first

    ...but I haven't been able to reproduce this issue.  Am I getting lucky and is `intptr_t` is being defined by some other header?  Is there something unique to your build configuration?  (Note, this is baselined against a different branch with CI config updates, but this shouldn't affect the Windows testing.)

2. Assuming that we do decide to explicitly include `stdint.h`, can you please move the include to `cl_platform.h`?  It could either be included in the Windows-specific section (OK) or we can pull it out of any OS-specific sections (better) since we'll be including it for all OSes for now.

Thanks again!

johnplate:
> Hi @johnplate , thank you for continuing to submit PRs to fix issues like these!

My pleasure :)

>    ...but I haven't been able to reproduce this issue.  Am I getting lucky and is `intptr_t` is being defined by some other header?  Is there something unique to your build configuration?

The CL headers are now in a project with really a lot of build configurations. Most were fine, but it broke our build on Windows with clang for arm64 target.

> 2. Assuming that we do decide to explicitly include `stdint.h`, can you please move the include to `cl_platform.h`? 

Sure, done. I have also documented the external dependency of the include position.

bashbaug (dismissed):

bashbaug:
> it broke our build on Windows with clang for arm64 target.

Ah, I see, I'm not sure that our Windows CI currently tests on arm64 so that's plausible.

Latest changes LGTM.  Thanks!

jrprice (commented):
Presumably the reason we had this include inside an "if not Windows" conditional is because there are older versions of MSVC that do not provide it. A quick search suggests that it was missing prior to Visual Studio 2010.

Do we still care about these older build systems? If so, this is not the right fix.

On the other hand, if we do _not_ still care about them, then we should go further and remove many of the conditional typedefs here.

johnplate:
James is right, stdint.h was missing in Visual Studio 2005 and 2008. It would be great if the supported build configurations would be documented somewhere, so we could determine how to proceed.

bashbaug:
Good point.  I suppose we could check the MSVC version using `_MSC_VER` and only include `stdint.h` for versions of Visual Studio > 2010?

https://docs.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=msvc-160

jrprice:
> It would be great if the supported build configurations would be documented somewhere

Yep. We have issue #127 to track this, but we don't seem to have produced a conclusion yet.

jrprice:
> suppose we could check the MSVC version using `_MSC_VER` and only include `stdint.h` for versions of Visual Studio > 2010?

I'm not sure that's enough to get this building on the older MSVCs though. Where would they get their definition of `intptr_t` from? Do we need to introduce a `cl_intptr` and then either `typedef __ptr cl_intptr` or `typedef intptr_t cl_intptr` depending on the platform?

Presumably the headers don't build on MSVCs older than 2010 because they use intptr_t, and probably haven't for a while. If nobody has noticed, maybe that's a good indicator that we could set C99 as a minimum supported standard?

bashbaug:
I think it's more that the headers from earlier versions of MSVC are picking up `intptr_t` from somewhere else, so they don't need to explicitly include `stdint.h`, which is good because it doesn't exist. 😄 

In case you missed it, I haven't been able to reproduce this problem locally or via our CI, so we're probably just getting lucky on most of the Windows build configurations.

https://github.com/bashbaug/OpenCL-Headers/tree/include-cl-header-first

johnplate:
Visual C++ provides intptr_t in stddef.h since forever, that's why Ben couldn't reproduce the bug. I have changed it now to only include stdint.h on Windows if clang is used or _MSC_VER >= 1600 (Visual Studio 2010)

jrprice (approved):
> Visual C++ provides intptr_t in stddef.h since forever

Thanks, that was the missing piece of the puzzle for me.

bashbaug (approved):
Thanks!

PR #179, opened 2021-10-06
Fix condition for warning 4201 pop

y-novikov:
Matching the condition for push.
New clang compiler version complains that the warning can be popped
without being pushed first.

### DIFF
--- a/CL/cl_platform.h
+++ b/CL/cl_platform.h
@@ -1395,7 +1395,9 @@ typedef union
 }
 #endif
 
-#if defined( _WIN32) && defined(_MSC_VER) && ! defined(__STDC__)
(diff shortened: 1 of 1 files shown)


y-novikov:
Fails when compiling ANGLE with a newer clang version here:
https://ci.chromium.org/ui/p/angle/builders/try/win-clang-x64-rel/23818/overview

Kerilk (approved):
This looks like the most maintainable way to do it. Thanks for the contribution.

If we want to avoid this issue in the future maybe we should add a comment to update both conditional flows somewhere.

MathiasMagnus (approved):
Also looks like a good change.

PR #182, opened 2021-11-12
Add provisional command-buffer extension

EwanC:
Run `gen_headers.py` from https://github.com/KhronosGroup/OpenCL-Headers/pull/161 on XML file containing https://github.com/KhronosGroup/OpenCL-Docs/pull/711 new spec extensions.

The only XML changes are for the `cl_khr_command_buffer` extension, which declarations get generated at the top of `cl_ext.h` by the script. So I have taken the generated `cl_khr_command_buffer` code and applied it to the same location in master branch `cl_ext.h`.

### DIFF
--- a/CL/cl_ext.h
+++ b/CL/cl_ext.h
@@ -38,8 +38,9 @@ typedef struct _cl_command_buffer_khr* cl_command_buffer_khr;
 typedef cl_uint             cl_sync_point_khr;
 typedef cl_uint             cl_command_buffer_info_khr;
 typedef cl_uint             cl_command_buffer_state_khr;
-typedef cl_bitfield         cl_command_buffer_properties_khr;
-typedef cl_bitfield         cl_ndrange_kernel_command_properties_khr;
+typedef cl_properties       cl_command_buffer_properties_khr;
+typedef cl_bitfield         cl_command_buffer_flags_khr;
+typedef cl_properties       cl_ndrange_kernel_command_properties_khr;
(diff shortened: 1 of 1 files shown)


bashbaug (approved):
LGTM, though we should probably hold off merging this until the spec PR is merged, just in case there are some last-minute changes.

My one minor comment is that this PR includes two "features" from the codegen scripts that I think are good but we haven't formally agreed to:

* It includes a define for the extension name `CL_KHR_COMMAND_BUFFER_EXTENSION_NAME`.
* It include the `CL_NO_PROTOTYPES` guard around the function prototypes.

I'm fine including these now, though if we want to stay consistent with the other extensions in this header at least until the codegen changes are merged we could remove them for now.

Thanks!

bashbaug (changes requested):
Needs to be updated to remove "INFO" from the names of the command buffer query functions, see: https://github.com/KhronosGroup/OpenCL-Docs/pull/711#pullrequestreview-805504564

bashbaug (approved):

bashbaug (approved):

alycm (approved):
> declarations get generated at the top of cl_ext.h by the script. So I have taken the generated cl_khr_command_buffer code and applied it to the same location

It looks like the generation script goes alphabetically.  No functional difference, so I don't mind where it goes.

> My one minor comment is that this PR includes two "features" from the codegen scripts that I think are good but we haven't formally agreed to:

These seem harmless.

bashbaug:
Spec changes are merged so I think we can merge this now, too.  Thanks!

PR #183, opened 2021-11-13
Rename cl_intel_thread_local_exec to cl_intel_exec_by_local_thread

fzou1:
cl_intel_exec_by_local_thread is the correct extension name.


bashbaug (approved):
LGTM.  There is a very slight chance this could break shipping code if it is looking for the name of the old define,  but I'm guessing we're OK with that?

bashbaug:
I'll go ahead and merge this.  If the changed define is a problem we can revisit.  Thanks!

fzou1:
Thank you.

PR #187, opened 2021-11-24
Add definitions for cl_arm_protected_memory_allocation

kpet:
Specification: https://github.com/KhronosGroup/OpenCL-Docs/pull/727

### DIFF
--- a/CL/cl_ext.h
+++ b/CL/cl_ext.h
@@ -1684,7 +1684,7 @@ typedef cl_uint cl_command_termination_reason_arm;
 
 #define cl_arm_protected_memory_allocation 1
 
-#define CL_MEM_PROTECTED_ALLOC_ARM (1 << 36)
+#define CL_MEM_PROTECTED_ALLOC_ARM (1ULL << 36)
 
 /******************************************
 * cl_intel_exec_by_local_thread extension *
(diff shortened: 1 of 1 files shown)


bashbaug (commented):
Some compilers will emit a warning and generate a poison value for this case: https://godbolt.org/z/dKo3KjThz

I think we'll need something else where we're shifting by more than 31 bits.  Some options:

* `((cl_ulong)1 << 36)` (cl_ulong is always 64-bits)
* `((cl_bitfield)1 << 36)` (cl_bitfield is also always 64-bits)
* `(1ULL << 36)` (long long is guaranteed to be 64-bits in C++11 at least, not sure about others)

I'll update the header generation scripts when we figure out what we want to do.

kpet:
Oh yes, it's undefined behaviour. I've gone for `ULL`. This should have been caught by the CI. Maybe we can use `Werror`. I'll have a look.

kpet:
@bashbaug Could you take another look please? We'd like to get this in ahead of an upcoming release. Thanks!

bashbaug (approved):
Sorry about the slow reply - this got lost in my notifications. Latest version LGTM.  Merging!

PR #189, opened 2021-12-03
OpenCL-SDK base changes

MathiasMagnus:
This PR supports OpenCL-SDK changes. The PR contains:

- GitLab CI sample which people can draw inspiration from to test the OpenCL-Headers repo on non-GA runners.
  - The images which the GitLab CI YAML uses can be found in in the OpenCL-SDK this PR supports.
- Updated REAMDE.md detailing the building and depending on the Headers using CMake.
  - A complete guide on CMake build system support is being merged into the OpenCL-Guide via [this](https://github.com/KhronosGroup/OpenCL-Guide/pull/16) PR.


CMakeLists.txt
```diff
@@ -1,5 +1,10 @@
 cmake_minimum_required(VERSION 3.0)
 
+# Include gudard for including this project multiple times
```
bashbaug, line 3: Typo fix:

```suggestion
# Include guard for including this project multiple times
```

.gitlab/stream-ci.yaml
```diff
@@ -0,0 +1,199 @@
...
+    expire_in: 2 weeks
+
+
+################
+##            ##
+## Test stage ##
+##            ##
+################
+
+## ROCm
```
bashbaug, line 104: Is this a typo or a copy-paste issue?  I don't see anything here having to do with ROCm.

bashbaug (commented):
Mostly LGTM.  I'm admittedly not entirely following what the CI changes are doing and these jobs don't seem to be run by default.  Do we need to change some repo settings to run them or will they start to run automagically when this PR is merged?

.gitlab/stream-ci.yaml
MathiasMagnus: Indeed, copy-pasta error from our own scripts. Removed to avoid confusion.

CMakeLists.txt
MathiasMagnus: Fixed.

MathiasMagnus:
@sl1pkn07 I can't seem to find the issue where you challanged the necessity of `SYSTEM BEFORE`. As it turns out (now that I revisited the issue), that was a desperate attempt at making MacOS happy, so that the C++ bindings pick up our headers and not the system ones.

Install rules have been added specific to MacOS that hide the system headers when our package is used. No more flaky `SYSTEM BEFORE` is required to control precedence of header paths. (Note that XCode and Ninja behave differently in this regard, as XCode adds a whole bunch of args to the command-line, so it was tricky to find one silver bullet that covers all use cases. And some are still not fully covered, but we'll improve CI coverage over time.)

sl1pkn07:
you mean on this https://github.com/KhronosGroup/OpenCL-ICD-Loader/pull/115 ?

MathiasMagnus:
I don't see SYSTEM BEFORE there, but in one incarnation of the CMake Packae Config support it was there and it was because of MacOS. This PR supercedes that series of PRs. The SDK samples base changes all have Package Config support and more fleshed out CI.

sl1pkn07:
sorry, i'm completely lost. I think you mistaken of person

i'm never talking about of `SYSTEM BEFORE`. but i have reviews pending about other things in KhronosGroup/OpenCL-ICD-Loader#115 and https://github.com/KhronosGroup/OpenCL-Headers/pull/107 (merged)

greetings

MathiasMagnus:
The [OpenCL-ICD-Loader case changes PR](https://github.com/KhronosGroup/OpenCL-ICD-Loader/pull/160) is looking ready, the last thing which needs attending to is remove the [temporary redirection](https://github.com/KhronosGroup/OpenCL-ICD-Loader/pull/160/files#diff-21364b2e6fae1f2875cee1ab3daefb0685403687eaf8bc32b5c6eacda351c9d3R31-R33) to this PR's source branch. This PR needs to go through before the ICD base changes can.

This already received a LGTM from Ben, I addressed his few concerns and fixed MacOS installation. I think it's good to go, but please take a look.

CMakeLists.txt
```diff
@@ -1,5 +1,10 @@
 cmake_minimum_required(VERSION 3.0)
 
+# Include guard for including this project multiple times
+if(TARGET Headers)
+  return()
+endif()
+
 project(OpenCLHeaders
   VERSION 2.2
```
Kerilk, line 9: Is this version correct? Might not be related to this PR, or is it? I am no CMake expert, but has the behavior of the CMake package changed significantly enough to warrant a package version change? Or should the version match those of the headers?

MathiasMagnus: This specific PR hasn't changed much, so I don't think so. I think bumping this version should be the subject of a separate discussion. IMHO it should be 3.0.x, but it should be discussed with the group.

.gitlab/stream-ci.yaml
```diff
@@ -0,0 +1,195 @@
+stages:
```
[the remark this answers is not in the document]
MathiasMagnus: Done.

CMakeLists.txt
Kerilk: Thanks for the insight.

Kerilk (approved):

PR #193, opened 2022-01-14
Add cl_intel_mem_alloc_buffer_location property

sherry-yuan:
This property allows user to specify which device global memory should be allocated to.

Supported by OpenCL 3.0 clCreateBufferWithProperties functions and OpenCL 1.x 2.x clCreateBufferWithPropertiesINTEL extension function.

Done to align with the spec in https://github.com/KhronosGroup/OpenCL-Docs/blob/master/extensions/cl_intel_mem_alloc_buffer_location.asciidoc

CC: @mkinsner @bashbaug @GarveyJoe

### DIFF
--- a/CL/cl_ext.h
+++ b/CL/cl_ext.h
@@ -2347,6 +2347,19 @@ clEnqueueMemsetINTEL(
 
 #endif /* CL_NO_PROTOTYPES */
 
+/***************************************************************
+* cl_intel_mem_alloc_buffer_location
+***************************************************************/
+#define cl_intel_mem_alloc_buffer_location 1
+#define CL_INTEL_MEM_ALLOC_BUFFER_LOCATION_EXTENSION_NAME \
+    "cl_intel_mem_alloc_buffer_location"
+
+/* cl_mem_properties_intel */
+#define CL_MEM_ALLOC_BUFFER_LOCATION_INTEL                  0x419E
+
+/* cl_mem_alloc_info_intel */
+/* enum CL_MEM_ALLOC_BUFFER_LOCATION_INTEL */
+
 /***************************************************
 * cl_intel_create_buffer_with_properties extension *
 ****************************************************/


CL/cl_ext.h
```diff
@@ -2347,6 +2347,19 @@ clEnqueueMemsetINTEL(
 
 #endif /* CL_NO_PROTOTYPES */
 
+/***************************************************************
+* cl_mem_alloc_buffer_location_intel
```
EwanC, line 2351: Extension name is `cl_intel_mem_alloc_buffer_location` not `cl_mem_alloc_buffer_location_intel`  as these defines suggest

CL/cl_ext.h
```diff
@@ -2347,6 +2347,19 @@ clEnqueueMemsetINTEL(
...
+***************************************************************/
+#define cl_intel_mem_alloc_buffer_location 1
+#define CL_INTEL_MEM_ALLOC_BUFFER_LOCATION_EXTENSION_NAME \
+    "cl_intel_mem_alloc_buffer_location"
+
+/* cl_mem_properties_intel */
+#define CL_MEM_ALLOC_BUFFER_LOCATION_INTEL                  0x419E
+
+/* cl_mem_alloc_info_intel */
+/* enum CL_MEM_ALLOC_BUFFER_LOCATION_INTEL */
```
EwanC, line 2361: I don't think this commented out line adds anything to the headers. Running the header generation script on cl.xml I see `CL_MEM_ALLOC_BUFFER_LOCATION_INTEL` getting defined twice, which i'm guessing is why this commented is here.

I think in https://github.com/KhronosGroup/OpenCL-Docs/blob/master/xml/cl.xml#L6409-L6414 we might have only wanted a single `<require>` entry for `CL_MEM_ALLOC_BUFFER_LOCATION_INTEL`

sherry-yuan: Thanks! When I ran the generation script, the line is automatically commented. I think the reason why there were 2 `<require>` in cl.xml is: this enum is indeed a valid property for both types (`cl_mem_alloc_info_intel ` and `cl_mem_properties_intel`). If I am understanding you correctly, by single require yo u mean the following?:
```
        <extension name="cl_intel_mem_alloc_buffer_location" supported="opencl">
            <require>
                <type name="CL/cl.h"/>
            </require>
            <require comment="cl_mem_properties_intel, cl_mem_alloc_info_intel">
                <enum name="CL_MEM_ALLOC_BUFFER_LOCATION_INTEL"/>
            </require>
        </extension>
```

EwanC: Apologies, I've checked out a more recent version of the header generation script and it does indeed match what you have :)

The XML you've posted was what I had in mind, but I'm not knowledgable enough about the XML to say if it's the right thing to do, or worth making a change over.

So change as it is LGTM

EwanC (approved):

CL/cl_ext.h
sherry-yuan: Ahh I see. Glad it is resolved, thanks for reviewing!

P.S I believe this format is what Ben had in mind, given he wrote up the generation script, I believe that should be the right format (or at least closest to correct).

bashbaug: Yes, I think everything is working as expected here (phew!).

I added "duplicate detection" to the header generation script a while back for two reasons:

1. For cases like this one, where the same enum might have two uses.
2. For extensions that share types or enums.

We can't generate the type or enum more than once so I thought it made sense to generate both but to "comment out" the latter ones so it's clear that they would have been generated if they weren't duplicates.

I think the XML file could be updated so the enum is only in one `<require>` entry if that'd make more sense, but this isn't required.

bashbaug (approved):
LGTM, I'll merge as soon as the CI results come back.

CL/cl_ext.h
oddhack: @bashbaug the schema is intended to support requiring the same API multiple times. The scripts we use in Vulkan / OpenXR / OpenVX / etc. do all the dependency tracking to ensure that APIs are emitted only once no matter how often they're referred to - in fact that's a big part of the script logic.

oddhack: N.b. there are a lot of odd corner cases that can be gotten into that are handled by my scripts and you may encounter eventually.

EwanC: Thanks for the background 👍

sherry-yuan: I see, thanks for the info!

PR #195, opened 2022-02-06
Option to disable tests when subproject

MathiasMagnus:
This PR follows the path paved by OpenCL-ICD-Loader crafted by @bashbaug [here](https://github.com/KhronosGroup/OpenCL-ICD-Loader/pull/117).

The behavior is:
- When built as top-level project
  - Tests by default are built.
  - To opt-out of testing, one uses the canonical `BUILD_TESTING` (on-by default)
- When built as a subproject
  - Tests are _not_ built by default
  - If tests are otherwise enabled (using the canonical `BUILD_TESTING` variable) one can opt-in to testing via `OPENCL_HEADERS_BUILD_TESTING`


bashbaug (approved):
LGTM, thanks!

PR #198, opened 2022-04-14
Fix warnings on anonymous structs for MSVC

MathiasMagnus:
The Windows CI runner has been updated and it broke our CI for multiple reasons boiling down the VS being updated within the image:

- Ninja runs broke because the location of `vcvarsall.bat` along with the VS install has changed.
- MSVC toolset v140 is no longer present as part of a VS 2022 install.
- v142 and v143 toolsets have _finally_ learned proper C11 and C17 conformance. This new support broke our macro detection logic turning off warnings of anon structs.

There are a few more updates I'd like to do (but time won't allow), so for the time being I fixed Windows CI definition and updated the headers to not break.


MathiasMagnus:
@Kerilk I don't think it does. MSVC has had anon structs since VS 2008 (toolset v90) and since it only warns about the use of anon structs, but never fails compilation, even in the new strict conformance modes, `/std:c11` and `/std:c17`, it shouldn't break existing workflows. I can't find any `/permissive-` analog from the C++ compilation mode, that switches between strict vs. with MS extensions modes, only `/Za` does that for C89 specifically. Using the new C lang specifiers seem to imply strict conformance. With this in mind, disabling the warning unconditionally should only break people using < VS 2008, which shouldn't be an issue. People rocking 17 year old compilers should fetch older headers. MSVC users have to make do with C11, C17 modes, and whatever version/extension soup the default flags give.

Having that said, one thing I noted is that we never test consuming the headers from C++ compilers, even though we specifically trigger header logic for anon structs depending on `__cplusplus__` or not, including the disabling of the warning. I've tweaked tests definition to run them through C++ compilers as well and I'm fixing warnings. The C headers should be parsed clean in C99, C11, C17, C++11, C++17 using `-Wall -Wextra -Werror` and `/W4 /WX`. (`-Weverything` and `/Wall` are still too much)

Kerilk:
@MathiasMagnus Sounds good, thanks for doing this.

bashbaug (commented):
These changes mostly look good to me and I like that the conditions are a bit simpler, but it does look like the logic is slightly different after this change and I'd like to confirm the difference is intentional.   The specific scenario is when the first test is false:

```c
#if !defined(__cplusplus) && defined(__STDC_VERSION__) && __STDC_VERSION__ >= 201112L
```

... and the next test is false:

```c
#elif defined( __GNUC__) && ! defined( __STRICT_ANSI__ )
```

Previously, we would fall into a third condition:

```c
#elif defined( _WIN32) && defined(_MSC_VER) && ! defined(__STDC__)
    #if _MSC_VER >= 1500
   /* Microsoft Developer Studio 2008 supports anonymous structs, but
    * complains by default. */
    #define  __CL_HAS_ANON_STRUCT__ 1
    #define  __CL_ANON_STRUCT__
   /* Disable warning C4201: nonstandard extension used : nameless
    * struct/union */
    #pragma warning( push )
    #pragma warning( disable : 4201 )
    #endif
```

We don't anymore though, and instead we go directly into the catch-all case where anonymous structures are not supported:

```c
#else
#define  __CL_HAS_ANON_STRUCT__ 0
#define  __CL_ANON_STRUCT__
#endif
```

With the missing third condition does it mean that some cases that enabled anonymous structures previously would now have anonymous structures disabled?

MathiasMagnus:
@bashbaug Yes, the changes were intentional, but ultimately were faulty, as they sometimes disabled anon structs when it wasn't strictly necessary. (Namely in the default MSVC case when you ask for the non-standard version soup.) Initially CI broke because the need for the warning disabling on Windows now became necessary on two branches, the 1st happy path and the MSVC version soup branch. To not duplicate the pragma stuff, I lifted it outside the version and dialect detection.

Note that I pimped the CI by breaking platforms into their own files and borrowing the OpenCL-Layers definition (which is an improvement over the OpenCL-SDKCI definition). CI improvements seem to propagate in a circular fashion.

tests/test_cl_half.h.c
```diff
@@ -95,7 +95,7 @@ int main(void)
   CHECK_FROM_FLOAT(-(CL_HALF_MAX + 1000.f), 0xfbff, RTP);
   CHECK_FROM_FLOAT(-(CL_HALF_MAX + 1000.f), 0xfc00, RTN);
   CHECK_FROM_FLOAT(-(CL_HALF_MAX + 1000.f), 0xfbff, RTZ);
-#ifndef _MSC_VER
+#if 0 // Hexadecimal float constant is C++17
```
Kerilk, line 98: This has been introduced in C99 and I checked that it is supported in C11

Kerilk (changes requested):
Looks good aside from the hexadecimal float constant issue.

tests/CMakeLists.txt
```diff
@@ -1,14 +1,29 @@
...
+  if(LANG_DIR MATCHES cpp)
+    set(CPP_SOURCE_PATH "${CMAKE_CURRENT_BINARY_DIR}/${SOURCE}pp")
+    file(RELATIVE_PATH REL_SOURCE_PATH "${CMAKE_CURRENT_BINARY_DIR}" "${CMAKE_CURRENT_SOURCE_DIR}/../${SOURCE}")
+    file(TO_NATIVE_PATH "${REL_SOURCE_PATH}" NATIVE_SOURCE_PATH)
+    file(WRITE
+      "${CPP_SOURCE_PATH}"
+      "#include \"${NATIVE_SOURCE_PATH}\"\n"
+    )
+    set(SOURCE_PATH "${CPP_SOURCE_PATH}")
+    set(LANG cpp)
```
bashbaug, line 17: Should we just add the .cpp files vs relying on cmake trickery to do this for us?

tests/test_cl_half.h.c
bashbaug: Are you OK merging this as-is to unblock CI?  We can fix this and any other minor issues in a subsequent PR.

Kerilk: I guess I would, but this removes tests that maybe are only done here, so we must not forget about it.

bashbaug (approved):
I filed issue https://github.com/KhronosGroup/OpenCL-Headers/issues/199 so we don't forget to re-enable testing for cl_half.h.

I still think we should add .cpp test files directly, but I also don't think this should hold up merging.

Kerilk (approved):

tests/CMakeLists.txt
MathiasMagnus: This was the order of things I tried:
- First, just add two executables in the same script file and for one of them tweak the source file lang. Source file languages however are controlled at directory level (`CMAKE_CURRNENT_LIST_DIR` level), so that changes it for both.
- Moving the scripts to separate directories and controlling the source lang that way introduces a warning on Clang saying [`warning: treating ‘A’ input as ‘B’ when in C++ mode, this behavior is deprecated`](https://clang.llvm.org/docs/DiagnosticsReference.html#wdeprecated) which can only be silenced bundled with other warnings and not selectively.
- Didn't know which was the better approach, to stamp out single-line files and double the number of files in the repo or generate them via CMake. Both are simple conceptually. One looks ugly in code, the other looks ugly on the command-line/File Explorer/Solution Explorer.

PR #200, opened 2022-05-24
Disable anon struct warnings for Clang on Win using GNU-like CLI

MathiasMagnus:
I started compiling OpenCL-Layers on Windows using clang using both `clang-cl.exe` and `clang.exe`. When using the GNU-like CLI, it warns about anon-structs like GCC does too and the `__extension__` decorator is required.


bashbaug (approved):
Works for me.  Do we need something similar for the `cl_image_desc` struct with the anonymous union, or is the existing Clang section in that struct which disables a warning sufficient?

```c
#ifdef __clang__
#pragma clang diagnostic push
#pragma clang diagnostic ignored "-Wc11-extensions" /* Prevents warning about nameless union being C11 extension*/
#endif
```

bashbaug:
This PR has been open for a while with no other reviews and CI is clean.  Merging.

PR #203, opened 2022-07-26
CMakeLists.txt: fix install on macOS if GNU coreutils are used

Mic92:
bsd ln supports both `-h` and `-n`, while GNU only supports `-n`.
In Nix we use GNU coreutils and hence the installation breaks.

Also see https://github.com/NixOS/nixpkgs/pull/182909

With this patch installation will work with both variants of `ln`


bashbaug (commented):
This looks reasonable to me.  I'd like a review from @MathiasMagnus or @jrprice before merging though.

MathiasMagnus (approved):
If CI is happy, I'm happy. I don't see anything wrong with the change.

PR #204, opened 2022-08-29
Update cl_ext.h for cl_intel_queue_no_sync_operations

aharon-abramson:
Adding macros for cl_intel_queue_no_sync_operations

### DIFF
--- a/CL/cl_ext.h
+++ b/CL/cl_ext.h
@@ -2441,10 +2441,14 @@ typedef struct _cl_queue_family_properties_intel {
 #define CL_QUEUE_CAPABILITY_BARRIER_INTEL                   (1 << 25)
 #define CL_QUEUE_CAPABILITY_KERNEL_INTEL                    (1 << 26)
 
-/* cl_intel_queue_no_sync_operations */
+/***************************************************************
+* cl_intel_queue_no_sync_operations
+***************************************************************/
+
 #define cl_intel_queue_no_sync_operations 1
 
-#define CL_QUEUE_NO_SYNC_OPERATIONS_INTEL                   (1 << 4)    
+/* addition to cl_command_queue_properties */
+#define CL_QUEUE_NO_SYNC_OPERATIONS_INTEL                   (1 << 27)    
     
 /***************************************************************
 * cl_intel_sharing_format_query
--- a/CL/cl_ext.h
+++ b/CL/cl_ext.h
@@ -2448,7 +2448,7 @@ typedef struct _cl_queue_family_properties_intel {
 #define cl_intel_queue_no_sync_operations 1
 
 /* addition to cl_command_queue_properties */
-#define CL_QUEUE_NO_SYNC_OPERATIONS_INTEL                   (1 << 27)    
+#define CL_QUEUE_NO_SYNC_OPERATIONS_INTEL                   (1 << 29)    
     
 /***************************************************************
 * cl_intel_sharing_format_query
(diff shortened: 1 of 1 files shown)


aharon-abramson:
@bashbaug, can you please review?

CL/cl_ext.h
```diff
@@ -2441,6 +2441,11 @@ typedef struct _cl_queue_family_properties_intel {
 #define CL_QUEUE_CAPABILITY_BARRIER_INTEL                   (1 << 25)
 #define CL_QUEUE_CAPABILITY_KERNEL_INTEL                    (1 << 26)
 
+/* cl_intel_queue_no_sync_operations */
+#define cl_intel_queue_no_sync_operations 1
+
+#define CL_QUEUE_NO_SYNC_OPERATIONS_INTEL                   (1 << 4)
```
bashbaug, line 2447: How much trouble would it be to choose a different bit number than "4" for this enum?

While there's no requirement to do this, for other bitfields (e.g. cl_mem_flags, cl_command_queue_properties) we've assigned bits for vendor extensions to higher bits (e.g. 31 on down), leaving the lower bits free for standard functionality.  I'd recommend continuing this convention if at all possible.

CL/cl_ext.h
```diff
@@ -2441,6 +2441,11 @@ typedef struct _cl_queue_family_properties_intel {
 #define CL_QUEUE_CAPABILITY_BARRIER_INTEL                   (1 << 25)
 #define CL_QUEUE_CAPABILITY_KERNEL_INTEL                    (1 << 26)
 
+/* cl_intel_queue_no_sync_operations */
```
bashbaug, line 2444: A few cosmetic suggestions:

* Can you expand this "header comment" to follow a similar format as the other extensions?  Right now it's easy to misinterpret these defines as a part of the preceding extension and not as a new extension.
* Can you add a comment describing which bitfield type `CL_QUEUE_NO_SYNC_OPERATIONS` is for?  I think it's an addition to `cl_command_queue_properties`?

We really should merge the header generation script (#161) so most of this will happen automatically...

bashbaug (changes requested):
Can you please also make a PR to update the XML grammar?

https://github.com/KhronosGroup/OpenCL-Docs/blob/main/xml/cl.xml

bashbaug:
Note: it would be best to have a non-Intel review approval before merging.

CL/cl_ext.h
aharon-abramson: Sure, no problem.

CL/cl_ext.h
```diff
@@ -2441,6 +2441,15 @@ typedef struct _cl_queue_family_properties_intel {
...
 #define CL_QUEUE_CAPABILITY_KERNEL_INTEL                    (1 << 26)
 
+/***************************************************************
+* cl_intel_queue_no_sync_operations
+***************************************************************/
+
+#define cl_intel_queue_no_sync_operations 1
+
+/* addition to cl_command_queue_properties */
+#define CL_QUEUE_NO_SYNC_OPERATIONS_INTEL                   (1 << 27)
```
EwanC, line 2451: why did you go for `27` out of interest? Wondering if there are other extensions that haven't reserved the xml value. `29` is the highest currently available bit position in https://github.com/KhronosGroup/OpenCL-Docs/blob/main/xml/cl.xml#L793

aharon-abramson: Yes, I got confused by CL_QUEUE_CAPABILITY_KERNEL_INTEL. I'll change it to 29.

CL/cl_ext.h
oddhack: There's a weak process note about reserving enums in cl.xml, but nothing about bits AFAIK. Making that process stronger (Vulkan has a very detailed set of steps to be taken before consuming shared resources like bits) doesn't guarantee no collisions or assigning values from the wrong end, but it definitely helps a lot.

EwanC (approved):

PR #208, opened 2022-09-27
require C99 for header tests

bashbaug:
See: https://github.com/KhronosGroup/OpenCL-SDK/issues/58#issuecomment-1259355455

Would requiring C99 explicitly help to resolve this issue?

If it does, should we require C99 for all parts of this project, or only for the tests?  Currently, this PR only requires C99 for the tests.


Kerilk:
My only question here derives from my lack of knowledge of cmake, and how projects including other cmake project (like the sdk or the loader) would react if they are build with other versions of the standards, but include this which specifies a precise version. @MathiasMagnus would know I assume.

MathiasMagnus:
@Kerilk

> ...how projects including other cmake project (like the sdk or the loader) would react if they are build with other versions of the standards...

It depends... (but I don't see a scenario where this could ever bite.) In the spirit of [teaching to fish](https://idioms.thefreedictionary.com/teach%20a%20man%20to%20fish), here's a full answer.

<on_cmake_variable_scoping>

The change proposed here sets the global defaults for language standards for targets that don't explicitly state what they should be using. It sets variables, and thus follows ordinary CMake variable scoping. (Functions and directories (meaning `add_subdirectory()`) introduce new scopes. Much like taking an arg by value in a function, even if something was set outside, changes to these global variables are only visible to things inside the directory and under. Setting it in the tests folder means that the change effects only the OpenCL Headers tests. It won't leak to consumers, no matter if they consume it through a package config file (an installation of the headers/SDK) or via making it a part of the their build (recursing into the source tree using `add_subdirectory()`). The `OpenCL::Headers` target was defined in one folder up.

<on_cmake_variable_scoping/>

<self_proclaimed_pro_tip>

If you don't like reasoning about global variables in C/C++ code, you probably don't like it in CMake either. It's possible to get it right (like here), but it's easy to get it wrong. Without going into a rant on all the occurences in the wild, I generally advise against "messing with the defaults". Modern CMake however favors using targets instead of variables. I personally would bump `cmake_minimum_required(VERSION 3.0)` to `3.1` (we already have that for the Loader, so it's hard to get around it anyway) and instead of setting the default for _all_ targets, inside `add_header_test` add it to the targets we intend to:
```cmake
set_target_properties(${TEST_EXE}
  PROPERTIES
    C_STANDARD 99
    C_STANDARD_REQUIRED ON
    CXX_STANDARD 11
    CXX_STANDARD_REQUIRED ON
)
```
These properties take their defaults from the similarly named `CMAKE_` prepended variables. If there is a convenient way to set the properties, I'd always go that way. I'd reserve using the variable form when targets are scattered throughout a multitude of folders. (And even then, I'd use an INTERFACE target, `OpenCLHeadersTestsCommon` and hook the property on that and link to it where needed.)

The rule of thumb I read from others, was that "anything starting with `CMAKE_` is for the user to set, not the project." In cases like lists, such as `CMAKE_PREFIX_PATH` and `CMAKE_MODULE_PATH` one may get away with appending/prepending to it, but it's a slippery slope.

<self_proclaimed_pro_tip/>

At the end of the day it's preference, and the proposed change given the strict relation to the properties it sets is equivalent to what I'd do.

juliohm:
Is this PR ready for review @bashbaug ?

Kerilk:
@MathiasMagnus Thanks a lot for the great answer!

bashbaug:
I found this one looking through old PRs.  Is this something we should pursue?  If not, I'll just close it.  @MathiasMagnus

MathiasMagnus (changes requested):
Minor change needed not to conflict with CI.

bashbaug:
@MathiasMagnus heads up, I made one change to your suggested fix, since I think it had a copy-paste error.  Otherwise, I think this one is ready to go.

MathiasMagnus (approved):
Ship it!

PR #209, opened 2022-09-27
use cinttypes instead of inttypes.h header for C++ tests

bashbaug:
see discussion: https://github.com/KhronosGroup/OpenCL-SDK/issues/58#issuecomment-1260074680

This change uses the `cinttypes` header when compiling C++ tests rather than the  C-based `inttypes.h` header.  Some versions of the `inttypes.h` header require also defining `__STDC_FORMAT_MACROS` to use the printf format macro constants used in the tests with C++ compiles, but not all.  Switching to the `cinttypes` header seems to work in all cases.

### DIFF
--- a/tests/test_headers.c
+++ b/tests/test_headers.c
@@ -1,5 +1,5 @@
 //
-// Copyright (c) 2020 The Khronos Group Inc.
+// Copyright (c) 2020-2022 The Khronos Group Inc.
 //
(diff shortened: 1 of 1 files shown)


Kerilk (approved):

juliohm:
Appreciate if someone can merge this approved PR 🙏🏽

PR #212, opened 2022-10-05
Formatted headers and added style check.

Kerilk:
This is a straw-man PR that formats and enable stye checking for the headers.

The reasoning behind this PR is thus:
 - Generating the headers from the xml specification is something the working group has been discussing for a long time;
 - Validating a newly generated set of headers would be difficult if sources cannot be easily compared;
 - Using an agreed upon style for the headers (eventually fixing inconsistencies) and enforcing it, would allow using the formatting tool on the generated headers;
 - This should allow easy validation of any change.

For now this PR uses clang-format using a custom style derived from Mozilla, which was the closest to ours. I assume it can still be refined. To apply the style to the headers run `clang-format -i -style=file **/*.h **/*.c` from the headers directory.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,149 @@
+---
+Language:        Cpp
+# BasedOnStyle:  Mozilla
+AccessModifierOffset: -2
+AlignAfterOpenBracket: Align
+AlignConsecutiveMacros: AcrossEmptyLinesAndComments
+AlignConsecutiveAssignments: false
+AlignConsecutiveBitFields: false
+AlignConsecutiveDeclarations: AcrossEmptyLinesAndComments
+AlignEscapedNewlines: Right
+AlignOperands:   Align
+AlignTrailingComments: true
+AllowAllArgumentsOnNextLine: true
+AllowAllConstructorInitializersOnNextLine: true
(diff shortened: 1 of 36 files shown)


bashbaug:
Regarding the specific style we use, in my opinion there should not be a strong requirement to match the formatting used by the current headers, especially because the the current formatting isn't always consistent.  As long as the style looks nice (yes, I know this is subjective) and is applied consistently it will be a readability win.

I do have a slight preference for styles that can be generated easily, although if we run clang-format as part of a generation script this is less important, also.

Kerilk:
> Regarding the specific style we use, in my opinion there should not be a strong requirement to match the formatting used by the current headers, especially because the the current formatting isn't always consistent. As long as the style looks nice (yes, I know this is subjective) and is applied consistently it will be a readability win.
> 
> I do have a slight preference for styles that can be generated easily, although if we run clang-format as part of a generation script this is less important, also.

Indeed, I started moving away from the style more and more. I would be very open to suggestion on the style to adopt. I am not opposed to radically changing the style, and adopt something used by other related communities.

Side note: based on my current experiments, there are a slight caveats with clang-format: it does not handle function pointers and function pointer typedefs extremely well.

PR #214, opened 2022-11-10
Added cl_icdl.h header file.

Kerilk:
This is a companion PR for:
https://github.com/KhronosGroup/OpenCL-ICD-Loader/pull/188


CL/cl_icdl.h
```diff
@@ -0,0 +1,46 @@
+/*
+ * Copyright (c) 2016-2020 The Khronos Group Inc.
```
Kerilk, line 2:
```suggestion
 * Copyright (c) 2016-2023 The Khronos Group Inc.
```

bashbaug (approved):
I'm fine with as-is, but just to ask the question: should we consider putting these additions into the existing cl_ext.h file vs. adding a new file?

I guess these extensions are little different than the other extensions in cl_ext.h: at least currently they won't go into any queryable extension strings, for example.  Perhaps that's a good enough reason to keep them separate?

Regardless, I think I'll try to update my header generation scripts to generate the defines and typedefs for this extension - seems like that should be pretty straightforward.

Kerilk:
Thanks for the insightful feedback.

> I'm fine with as-is, but just to ask the question: should we consider putting these additions into the existing cl_ext.h file vs. adding a new file?

Very good question. I discarded putting the extension in the `cl_icd.h` file as its usage is very specific and not application facing. I hadn't thought about `cl_ext.h`. Given users could want to invoke it from their applications it would make sense I guess. Tell me if you want me to go this route.

bashbaug (dismissed):

CL/cl_ext.h
```diff
@@ -2626,6 +2626,25 @@ clGetImageRequirementsInfoEXT_fn)(
...
+#define CL_ICDL_VERSION     2
+#define CL_ICDL_NAME        3
+#define CL_ICDL_VENDOR      4
+
+typedef cl_int
+(CL_API_CALL * pfn_clGetICDLoaderInfoOCLICD)(cl_icdl_info param_name,
+                                             size_t       param_value_size,
+                                             void *       param_value,
+                                             size_t *     param_value_size_ret);
+
```
bashbaug, line 2647: Noticed while adding support in the extension loader for this extension: I think it would be helpful to include a function prototype for this extension API also.  Something like:

```suggestion

#ifndef CL_NO_PROTOTYPES
extern CL_API_ENTRY cl_int CL_API_CALL
clGetICDLoaderInfoOCLICD(cl_icdl_info param_name,
                         size_t       param_value_size,
                         void *       param_value,
                         size_t *     param_value_size_ret);
#endif
```

This is essentially what the header generation scripts would generate.

Interestingly adding this does cause a build failure in the OpenCL ICD loader (I expect ocl-icd would have the same issue):

```sh
opencl-icd-loader/loader/icd_dispatch.c:28:1: error: static declaration of ‘clGetICDLoaderInfoOCLICD’ follows non-static declaration
```

We could fix this by defining `CL_NO_PROTOTYPES` to omit the function declaration, or we could make the function in the ICD loader non-static.... or we could rethink this entirely.  Let me know what you'd like to do.

Kerilk: I don't think we should declare `extern` symbols that the OpenCL library is not exporting.

I think we should provide ways for tools and users to easily declare the relevant entry points. This could come in the form of function typedefs. These could be used to create headers that would later be included (in you extension multiplexing library for instance) or to create an accurate `static` declaration as in the loader.

Ideally I would like to move to something akin to:

```C
typedef cl_int CL_API_CALL
clGetICDLoaderInfoOCLICD_t(cl_icdl_info param_name,
                           size_t       param_value_size,
                           void *       param_value,
                           size_t *     param_value_size_ret);

typedef clGetICDLoaderInfoOCLICD_t *pfn_clGetICDLoaderInfoOCLICD;
```

The user, or library can later declare it. Example in the loader
```C
static clGetICDLoaderInfoOCLICD_t clGetICDLoaderInfoOCLICD;
```
or in your extension library header:
```C
extern CL_API_ENTRY clGetICDLoaderInfoOCLICD_t clGetICDLoaderInfoOCLICD;
```

I know this is quite a departure from how we've been usually doing things, but I think it is the most flexible way. And it would lighten the headers. What do you think?

bashbaug: I think it's most important for the headers to stay internally consistent, so unless there is a very strong reason to deviate all extensions should be handled the same.  Since the current headers have an extern symbol for all extension APIs (with very few exceptions, namely for the DX sharing APIs, that is probably a bug) I think we should have an extern symbol for this extension API as well.

If we use `CL_NO_PROTOTYPES` to guard the extern symbols then this provides a mechanism for projects to opt-out and not declare extern symbols if they choose not too.  As an added bonus this also aligns with the Vulkan headers ([example](https://github.com/KhronosGroup/Vulkan-Headers/blob/main/include/vulkan/vulkan_core.h)) so we have some consistency across Khronos APIs as well.  We could discuss whether we should or should not have the extern symbols by default, but I personally think we'd want to err on the side of backwards compatibility and provide an opt-out mechanism not an opt-in mechanism.

I'd be happy to create a `CL_NO_PROTOTPES` PR for the ICD loader(s) if we decide to go this route.  I already have one for the OpenCL-CTS, though it hasn't been merged yet.

Kerilk: From a practical standpoint, this is fine with me for one caveat: we need to be careful as, some times, we will want to include these header with or without the flag. Hopefully this doesn't require too much reorganization in the loader to achieve. Especially, in Vulkan all declarations are guarded by this flag, so if our goal is to do the same, we have to make sure our strategy accounts for this being true at some point in the future.

> I think it's most important for the headers to stay internally consistent, so unless there is a very strong reason to deviate all extensions should be handled the same. Since the current headers have an extern symbol for all extension APIs (with very few exceptions, namely for the DX sharing APIs, that is probably a bug) I think we should have an extern symbol for this extension API as well.

I agree that DX should have declarations especially since they are exported by the loader. For the rest I understand your position but, for the record, I strongly disagree that we should be declaring non exported symbols at all.

Kerilk:
Closing as merged by https://github.com/KhronosGroup/OpenCL-Headers/pull/161

PR #219, opened 2023-02-17
install pkg-config if it is not already installed

bashbaug:
Looks like pkg-config is no longer available by default in the macos-11 image.  This PR checks to see if pkg-config is available and installs it if it isn't.  Needed to fix macOS CI.


Kerilk (approved):
LGTM. Will wait for CI and then merge.

PR #225, opened 2023-03-28
CI Fixes

Kerilk:
This fix two Issues we've been having recently withe the CI:
 - MacOS images are sometimes missing gcc-9, (seems to be transitioned out) so this checks for it and installs it if required.
 - Ubuntu 18.04 is deprecated, but as we run in a container we can safely update the Ubuntu version we are using so I bumped it to 20.04

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -4,7 +4,7 @@ on: [push, pull_request]
 
 jobs:
   compatibility:
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-20.04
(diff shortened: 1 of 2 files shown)


Kerilk:
> Since I think we'll need a similar fix elsewhere though (OpenCL-CLHPP, OpenCL-ICD-Loader, probably OpenCL-SDK), at some point we should investigate whether we can move from `macos-11` to `macos-latest` now that we're explicitly installing gcc9.

Indeed. I most probably will do one for the ICD loader in the process of validating https://github.com/KhronosGroup/OpenCL-ICD-Loader/pull/210

PR #231, opened 2023-05-22
fix the OpenCL version for the CL_DEPTH enum

### LINKED ISSUE #117: Depth Image Enums
See:

https://github.com/KhronosGroup/OpenCL-Docs/issues/284
https://github.com/KhronosGroup/OpenCL-Docs/pull/458

It looks like the enums for depth images in the headers have a similar problem:

```c
#ifdef CL_VERSION_1_2
#define CL_DEPTH                                    0x10BD
#define CL_DEPTH_STENCIL                            0x10BE
#endif
```

We should:

* Move some of these enums out of OpenCL 1.2 and into OpenCL 2.0 in cl.h, since this is when they became a core feature.
* Move / duplicate a few other enums into cl_ext.h, since they are extension enums and are not part of a core API.

bashbaug:
Fixes #117.

This change corrects the OpenCL version for the `CL_DEPTH` enum.  `CL_DEPTH` was an extension until OpenCL 2.0, not OpenCL 1.2, so it shouldn't be in the core `cl.h` until OpenCL 2.0.

Also fixes the extension enums `CL_DEPTH_STENCIL` and `CL_UNORM_INT24`, which were never core enums and hence should not be in the core `cl.h` at all.

The extension header fixes were made by regenerating headers using the XML from https://github.com/KhronosGroup/OpenCL-Docs/pull/930, so this change should not be merged until the XML changes are merged.

### DIFF
--- a/CL/cl.h
+++ b/CL/cl.h
@@ -569,11 +569,8 @@ typedef struct _cl_name_version {
 #define CL_RGx                                      0x10BB
 #define CL_RGBx                                     0x10BC
 #endif
-#ifdef CL_VERSION_1_2
-#define CL_DEPTH                                    0x10BD
-#define CL_DEPTH_STENCIL                            0x10BE
-#endif
 #ifdef CL_VERSION_2_0
+#define CL_DEPTH                                    0x10BD
 #define CL_sRGB                                     0x10BF
 #define CL_sRGBx                                    0x10C0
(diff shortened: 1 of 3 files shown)


Kerilk (approved):
LGTM. Release should include breakage scenario note.

bashbaug:
Merging as discussed in the September 12th teleconference, now that there is a review approval.

PR #233, opened 2023-06-20
Create deb package

Beanavil:
Adds a CMake module with the packaging configuration and "Package" and "Consume" steps to GitHub actions to check that it works as it should.

### DIFF
--- a/cmake/Package.cmake
+++ b/cmake/Package.cmake
@@ -1,0 +1,44 @@
+set(CPACK_PACKAGE_VENDOR "khronos")
+
+set(CPACK_PACKAGE_DESCRIPTION "OpenCL API C headers")
+
+set(CPACK_RESOURCE_FILE_LICENSE "${CMAKE_CURRENT_SOURCE_DIR}/LICENSE")
+
+set(CPACK_RESOURCE_FILE_README "${CMAKE_CURRENT_SOURCE_DIR}/README.md")
+
+if(NOT CPACK_PACKAGING_INSTALL_PREFIX)
+  set(CPACK_PACKAGING_INSTALL_PREFIX "${CMAKE_INSTALL_PREFIX}")
+endif()
+
+if(NOT CPACK_SOURCE_IGNORE_FILES)
+  set(CPACK_SOURCE_IGNORE_FILES="/.github/")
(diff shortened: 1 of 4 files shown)


cmake/Package.cmake
```diff
@@ -0,0 +1,41 @@
...
+endif()
+
+# DEB packaging configuration
+set(CPACK_DEBIAN_PACKAGE_MAINTAINER ${CPACK_PACKAGE_VENDOR})
+
+set(CPACK_DEBIAN_PACKAGE_HOMEPAGE
+    "https://github.com/KhronosGroup/OpenCL-Headers")
+
+# Version number [epoch:]upstream_version[-debian_revision]
+set(LATEST_RELEASE_VERSION "2023.04.17")
```
bashbaug, line 20: Out of curiosity where did this date come from?  Will it need to be updated manually?  If so, when should it be updated?

cmake/Package.cmake
```diff
@@ -0,0 +1,41 @@
...
+set(CPACK_DEBIAN_PACKAGE_VERSION "${PROJECT_VERSION}~${LATEST_RELEASE_VERSION}")  # upstream_version
+set(CPACK_DEBIAN_PACKAGE_RELEASE "1") # debian_revision (because this is a
+                                      # non-native pkg)
+set(PACKAGE_VERSION_REVISION "${CPACK_DEBIAN_PACKAGE_VERSION}-${CPACK_DEBIAN_PACKAGE_RELEASE}")
+
+set(DEBIAN_PACKAGE_NAME "opencl-c-headers")
```
bashbaug, line 26: Is this an existing package name or something new?  If this is a new package name I'd suggest removing "c", since I tend to think of "OpenCL C" as being the kernel language.

bashbaug (commented):
I'm not sure how to test these changes and I'm not a packaging expert, but this all looks reasonable to me.  I think the odds of breaking any existing functionality is low, therefore we should merge this sooner rather than later and we will perhaps get more feedback.

Beanavil:
Closing this, as the functionality was added in https://github.com/KhronosGroup/OpenCL-Headers/pull/239

PR #236, opened 2023-08-21
sort new extensions programmatically

bashbaug:
The generated extension headers maintained an ordered list of older extensions to minimize diffs with the non-generated extension headers, but there is no reason to require this for new extensions.  If a new extension is not found in the list, sort it at the end automatically. Note, Khronos (KHR) extensions will also be sorted before multi-vendor (EXT) extensions, and before vendor extensions.

### DIFF
--- a/scripts/cl_ext.h.mako
+++ b/scripts/cl_ext.h.mako
@@ -162,7 +162,16 @@ orderedExtensions = [
 
 def getExtensionSortKey(item):
     name = item.get('name')
-    return orderedExtensions.index(name)
+    try:
+        index = orderedExtensions.index(name)
+    except ValueError:
+        if name.startswith('cl_khr'):
+            index = 10000
+        elif name.startswith('cl_ext'):
+            index = 10001
+        else:
+            index = 10002
+    return index, name
 
 # Order the extensions should be emitted in the headers.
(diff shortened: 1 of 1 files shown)


scripts/cl_ext.h.mako
```diff
@@ -162,7 +162,16 @@ orderedExtensions = [
 
 def getExtensionSortKey(item):
     name = item.get('name')
-    return orderedExtensions.index(name)
+    try:
+        index = orderedExtensions.index(name)
+    except ValueError:
+        if name.startswith('cl_khr'):
+            index = 10000
```
kpet, line 169: What's the behaviour when the same index is returned multiple times? Are definitions for extensions ordered following the order in which they appear in the XML?

bashbaug: I had to look this up.  The sort is stable if they have the same key:
https://docs.python.org/3/howto/sorting.html#sort-stability-and-complex-sorts

However, the sort key is actually a tuple consisting of the index and the name, so the end result should be that the new extensions are ordered after the older manually ordered extensions and alphabetized.  This is the order I'd like to see for all extensions eventually - see `getExtensionSortKey_ideal` below.

kpet (approved):

PR #238, opened 2023-09-08
add compatibility notes

bashbaug:
Adds a short section to the README to describe the OpenCL headers compatibility policy.  In general, we will strive to maintain backward compatibility, but in rare cases we may break backward compatibility, especially for provisional features or extensions.

Also updates the CI scripts to skip GitHub actions if only README files are updated.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -1,6 +1,12 @@
 name: Linux
 
-on: [push, pull_request]
+on:
+  push:
+    paths-ignore:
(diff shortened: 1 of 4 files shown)


bashbaug:
Discussed in the September 12th teleconference.  We should add or update the existing provisional extension text in the spec, also.

EwanC (approved):

bashbaug:
Merging as discussed in the September 26th teleconference.

PR #246, opened 2024-02-17
Fix CMake deprecation warning

s-barannikov:
Recent versions of CMake issue a deprecation warning:
```
CMake Deprecation Warning at CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 3.5 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.
```

Silence this warning by using a `...<max>` suffix
in `cmake_minimum_required` invocation.
No functional change intended.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,4 @@
-cmake_minimum_required(VERSION 3.0)
-cmake_policy(VERSION 3.0...3.22)
+cmake_minimum_required(VERSION 3.0...3.22)
 
 # Include guard for including this project multiple times
 if(TARGET Headers)


MathiasMagnus (approved):
It's a good change.

bashbaug:
@MathiasMagnus should we be making similar changes to any of our other repos (e.g. OpenCL ICD loader, OpenCL SDK)?  If so, how do you want to track this?

MathiasMagnus:
Our CI/CD changes sport a similar change (and will conflict with this one, but we'll fix), so the other repos will get similar treatment without any special care.

Issue #248, opened 2024-03-19
Provide version macros for individual extensions


kpet:
We should provide macros for the version of individual extensions so applications can more easily adapt to changes to definitions for extensions (backwards compatible or not).

EwanC:
Want to confirm I understand this proposal because I agree having clear messaging to users on how to deal with API changes is a nice improvement.

The goal is that users will be able to write code like this?

```cpp
// Application code.
#if CL_FOO_KHR_VERSION == 1
  clFooKHR(bar);
#elif CL_FOO_KHR_VERSION == 2
  clFooKHR(baz);
#endif
```

And in `cl_ext.h` we'd define the macro like:
```cpp
// cl_ext.h
#define CL_FOO_KHR_VERSION N
// APIs for revision N ....
// No APIs for revisions earlier than N
```

And ideally the version macro would be generated from the `revision` component in the XML `<extension>` entry using `gen_headers.py`.

 Is that what you had in mind?

oddhack:
It seems like you would need a runtime query for the actual extension version supported by the driver, rather than statically compiling against the version that happens to be in the header file you built the app with?

kpet:
@EwanC That's exactly what I had in mind.

@oddhack We already have runtime queries  but we're missing a mechanism for compile-time checks similar to the `*_SPEC_VERSION` macros that the Vulkan headers provide. It's particularly problematic for provisional extensions that are less stable and for which breaking changes can occur between released versions of the headers.

oddhack:
FWIW, the Vulkan experience led to our current approach where we disallow breaking changes in extension revisions - they are only to indicate fixes to spec language, not behavior. If a shipping extensions is fundamentally broken then we may do a new extension that replaces or modifies it, which is thankfully infrequent.

There are two exceptions I can think of offhand:

- Occasionally a vendor ships a new vendor extension and almost immediately realizes something was wrong with it. Since the burden is entirely on that vendor, if they want to make a breaking change that's their call.
- Provisional extensions can change in an arbitrary fashion and as you say, we do have the compile-time macro to rely on, but not the runtime query as you do. The burden is on the ISV to ensure they are using a driver consistent with their header, which has not led to complaints so far (or at least, not that I'm aware of - vendors may be getting negative feedback we don't hear about).

kpet:
[...]

PR #251, opened 2024-05-07
Introduce extension macros

EwanC:
Implement idea from https://github.com/KhronosGroup/OpenCL-Headers/issues/248 to add an version macro to the extension headers so that users can guard application code using the macro to ensure the correct APIs are being used. This uses the XML from https://github.com/KhronosGroup/OpenCL-Docs/pull/1161 which forces extension to set a revision value.

Extensions can then increment the version when they change the APIs without breaking user code, providing applications have guarded their usage.

See [CL_MAKE_VERSION](https://registry.khronos.org/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_MAKE_VERSION) for how the macro version is defined. Note that the definitions of this macro and `cl_version` have been taken out of the  `CL_VERSION_3_0` version guard to allow them to always be used by an application. This is because there functionality is implemented in the headers and not affected in any way by the OpenCL implementation being used, and the OpenCL version that implementation supports.


CL/cl_ext.h
```diff
@@ -1776,21 +1903,12 @@ clEnqueueReleaseExternalMemObjectsKHR(
 #define CL_KHR_EXTERNAL_MEMORY_DMA_BUF_EXTENSION_NAME \
     "cl_khr_external_memory_dma_buf"
 
-/* cl_external_memory_handle_type_khr */
```
EwanC, line 1779: `cl_khr_external_memory_dx` has dissapeared after copying over `cl_ext.h` wholesale from the `gen_headers.py` output - However it looks like this was intended https://github.com/KhronosGroup/OpenCL-Docs/pull/1160

CL/cl_ext.h
```diff
@@ -1864,23 +1992,17 @@ clGetSemaphoreHandleForTypeKHR(
 
 #endif /* !defined(CL_NO_NON_ICD_DISPATCH_EXTENSION_PROTOTYPES) */
 
-/***************************************************************
-* cl_khr_external_semaphore_dx_fence
```
EwanC, line 1868: `cl_khr_external_semaphore_dx_fence` has dissapeared after copying over `cl_ext.h` wholesale from the `gen_headers.py` output - However it looks like this was intended https://github.com/KhronosGroup/OpenCL-Docs/pull/1160

bashbaug (approved):
LGTM.

I think `cl_ext.h` is going to change slightly when/if we remove the automated formatting checks - see https://github.com/KhronosGroup/OpenCL-Headers/pull/253 - but no need to hold up this PR until the automated formatting checks are merged.  It'll just make for a few more diffs next time we re-generate headers.

Un-guarding `CL_MAKE_VERSION` means that a pre-OpenCL 3.0 application that uses `CL_MAKE_VERSION` intentionally or unintentially (say, if it meant to use `CL_MAKE_VERSION_KHR` instead) would build with newer headers and fail to build with older headers, but the risk of that happening is probably low enough that we shouldn't worry too much about it.  I don't have a better solution, anyhow.

bashbaug:
Merging as discussed in the June 18th teleconference.

PR #252, opened 2024-05-17
Update CI matrix

mfep:
The list of compilers/environments is explicit in the presubmit CI pipeline. Since the `-latest` GitHub runners are subject to periodic change, the matrix also needs to be updated from time to time.

An alternative would be to pin the jobs at specific runner versions, but since those versions cease to be supported over time, the porting of jobs to a new runner would be a larger monolithic task instead of the current gradual transition.

This PR fixes the following:
1. Fixed the path of `gcc-11` and `gcc-13` compiler in the `macos` jobs.
2. Removed compilation against the Visual C++ v141 toolset, which is no longer available in the `windows-latest` installation.

### DIFF
--- a/.github/workflows/presubmit.yml
+++ b/.github/workflows/presubmit.yml
@@ -507,17 +507,17 @@ jobs:
       matrix:
         C_COMPILER:
         - /usr/bin/clang
-        - /usr/local/bin/gcc-11
-        - /usr/local/bin/gcc-13
+        - gcc-11
+        - gcc-13
         GEN:
         - Xcode
         - Ninja Multi-Config
         STD: [99, 11, 17]
         exclude:
         # These entries are excluded, since XCode selects its own compiler
-        - C_COMPILER: /usr/local/bin/gcc-11
+        - C_COMPILER: gcc-11
           GEN: Xcode
-        - C_COMPILER: /usr/local/bin/gcc-13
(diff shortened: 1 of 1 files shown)


MathiasMagnus:
@mfep you beat me to it. Thanks for the rapid response.

MathiasMagnus (approved):

MathiasMagnus (changes requested):
While we're at updating the CI matrix, one peculiarity of the C Headers repo was glossed over: that it does use C++ compilers to test for warning-free compilation. (It has bit in the past, that things compiled cleanly in C mode, but not in C++.) If one takes a look at the configure output of either Linux ([this](https://github.com/KhronosGroup/OpenCL-Headers/actions/runs/9126808716/job/25095835143?pr=252) or MacOS ([this](https://github.com/KhronosGroup/OpenCL-Headers/actions/runs/9126808716/job/25095827835?pr=252)) runs, it can be seen that when some compiler is favored on the platform (GCC on Linux, AppleClang on MacOS) then leaving the CXX compiler unspecified will result in mixed compiler usage. I believe the intention here was to test using the same toolchain for both C and CXX tests.

MathiasMagnus (approved):

bashbaug:
Merging as discussed in the May 21st teleconference.

Issue #254, opened 2024-06-18
fix CI failures


bashbaug:
Looks like some of our CI configurations are failing after an environment update that occurred sometime between May 21st (when the automated testing for #251 passed) and June 11th (when the automated testing for #253 failed).  Since the automated tests on the main branch are failing now too, these don't seem to be sporadic failures, and we need to fix or disable them.

bashbaug:
CC @Beanavil @mfep @MathiasMagnus

Beanavil:
The windows failure https://github.com/KhronosGroup/OpenCL-Headers/actions/runs/9472987725/job/26099412897?pr=253 is due to a breaking change in the MSVC versioning (see [cppblog](https://devblogs.microsoft.com/cppblog/whats-new-for-cpp-developers-in-visual-studio-2022-17-10/)):

> We have incremented the minor version number of the MSVC toolset from 19.39 (VS 2022 v17.9) to 19.40 (VS 2022 v17.10)

The fix was already added to the `Update CI, CD` PRs for other repos, for instance https://github.com/KhronosGroup/OpenCL-SDK/pull/88/commits/3a3b557f3a245c527354cf0a37dd3d883d8061f2

Beanavil:
For the MacOS one, I think this is a problem with the runner image itself (see [#9997](https://github.com/actions/runner-images/issues/9997)). Seems to be fixable with the suggestion in [this comment](https://github.com/actions/runner-images/issues/9997#issuecomment-2175693073), or by directly disabling `gcc-13/g++-13` from the CI matrix until the runner images are fixed (this was done in the [OpenCL-ICD-Loader/#226](https://github.com/KhronosGroup/OpenCL-ICD-Loader/pull/226/commits/95b2cf6a6bd5027b2e0345eef785228d0773693a)). I tested both approaches and both work:
- [Use compatible Xcode](https://github.com/Beanavil/OpenCL-Headers/actions/runs/9580712853)
- [Disable gcc-13/g++-13](https://github.com/Beanavil/OpenCL-Headers/actions/runs/9580455826)

bashbaug:
Fixed by #256 - thanks!

PR #256, opened 2024-06-25
Fix CI failures

Beanavil:
- Addressed #254.
- Updated CI docker images from StreamHPC's to KhronosGroup's.
- Updated CI actions to use Node.js 20.


bashbaug (approved):
Thank you for fixing these!

PR #257, opened 2024-07-10
Update generated headers and remove CL/GL types from cl_platform.h

kpet:
- Move CL/GL type declarations from cl_platform.h to cl_gl.h
- Include version macros for extensions

### DIFF
--- a/CL/cl_d3d10.h
+++ b/CL/cl_d3d10.h
@@ -65,6 +65,9 @@ extern "C" {
 #define CL_KHR_D3D10_SHARING_EXTENSION_NAME \
     "cl_khr_d3d10_sharing"
 
+
+#define CL_KHR_D3D10_SHARING_EXTENSION_VERSION CL_MAKE_VERSION(1, 0, 0)
+
 typedef cl_uint             cl_d3d10_device_source_khr;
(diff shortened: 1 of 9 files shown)


bashbaug (commented):
I'm OK with this and I think it's the right thing to do, but cl_gl.h is currently generated, so if we merge this as-is we'll need to be careful not to overwrite this change next time we re-generate headers.

If possible, I think it would be preferable to:

1. Make this change to cl.xml first, either as part of #960 or separately, if it'll take some time for #960 to land.
2. Regenerate headers with the updated XML file to update cl_gl.h.

Would this be OK?  I'll happily review and merge a PR for (1) as soon as it's ready.  Thanks!

kpet:
Oh yes, you're right. The few seconds I spent observing the absence of a mako template for `cl_gl.h` were not enough!

I've created a PR (https://github.com/KhronosGroup/OpenCL-Docs/pull/1210) separate from KhronosGroup/OpenCL-Docs#960 (that I agree will likely require a bit more discussion) and repurposed this one to update the generated headers and manually remove the types from `cl_platform.h`.

bashbaug (approved):

bashbaug:
I think this should be fairly non-contentious, so I'm going to go ahead and merge this offline.

PR #264, opened 2024-10-02
Update headers regenerated from latest xml

nikhiljnv:
Update OpenCL headers by regenerating them from
latest XML files to include recent spec updates.

### DIFF
--- a/CL/cl_ext.h
+++ b/CL/cl_ext.h
@@ -662,7 +662,7 @@ typedef cl_bitfield         cl_mutable_dispatch_asserts_khr;
 /* cl_device_info */
 #define CL_DEVICE_MUTABLE_DISPATCH_CAPABILITIES_KHR         0x12B0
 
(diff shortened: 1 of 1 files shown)


EwanC (approved):
LGTM - ran the script locally and this matches what I see

bashbaug (approved):

bashbaug:
Is this going to need updates to the CTS tests also, for the mutable command buffer changes?

EwanC:
> Is this going to need updates to the CTS tests also, for the mutable command buffer changes?

I think it's just the code comments that are changing here for mutable command-buffers

bashbaug:
> > Is this going to need updates to the CTS tests also, for the mutable command buffer changes?
> 
> I think it's just the code comments that are changing here for mutable command-buffers, so shouldn't be any effect on header users.

Ah right, thanks!  I'll just merge this, then.

PR #267, opened 2024-10-23
Temporarily disable PPA builds

MathiasMagnus:


bashbaug (approved):
I have no way to test this but it looks good to me.

Merging as we discussed in the October 22nd teleconference.

PR #268, opened 2024-11-06
guard provisional extensions with an opt-in define

### LINKED ISSUE #247: Guard definitions for provisional extensions with a dedicated opt-in macro
Many new extensions are being developed as provisional extensions. As part of their development the OpenCL WG may need to break backwards compatibility which may sometimes (has in the past) exposed applications that do not need provisional definitions to breaking changes. As discussed in the 2024/03/19, we should guard definitions for provisional extensions with an opt-in macro to limit exposure to breaking changes to those applications that use provisional extensions.

bashbaug:
fixes #247 

Wraps provisional extensions in an opt-in define so applications that do not use provisional extensions do not get broken by non-backwards compatible changes.

Notes:

* The define gets added automatically during header generation based on the `provisional="true"` line in the XML file.
* The define is currently an opt-in named `CL_PROVISIONAL_EXTENSIONS`, but it's trivial to change this if desired.
* Because it is an opt-in define, the current implementation does break backwards compatiblity, and applications using provisional extensions will need to add the opt-in define.
* We should consider how much additional testing we want to add (in this repo and in others) with this change.  For example, do we want to run some of the header tests with and without provisional extensions?

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -34,18 +34,31 @@ function(add_header_test NAME SOURCE)
     set(LANG c)
   endif()
   foreach(VERSION 100 110 120 200 210 220 300)
-    set(TEST_EXE ${NAME}_${LANG}_${VERSION})
-    list(FIND TEST_SKIP_ANSI_TESTING ${NAME} TEST_SKIP_INDEX)
-    if(NOT (${TEST_SKIP_INDEX} GREATER -1 AND MSVC AND CMAKE_C_FLAGS MATCHES "/Za"))
-      add_executable(${TEST_EXE} "${SOURCE_PATH}")
-      target_compile_definitions(${TEST_EXE}
-        PUBLIC -DCL_TARGET_OPENCL_VERSION=${VERSION}
-      )
-      target_include_directories(${TEST_EXE}
-        PUBLIC ${PROJECT_SOURCE_DIR}
-      )
-      add_test(NAME ${TEST_EXE} COMMAND ${TEST_EXE})
-    endif()
+    foreach(OPTION "" CL_ENABLE_PROVISIONAL_EXTENSIONS)
+      if(OPTION STREQUAL "")
+        # The empty string means we're not setting any special option.
+        set(UNDERSCORE_OPTION "${OPTION}")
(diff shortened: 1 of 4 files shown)


bashbaug:
Vulkan uses `VK_ENABLE_BETA_EXTENSIONS`, so I have a slight preference for "enable".

How closely do we want to align?  Should we use `CL_ENABLE_BETA_EXTENSIONS`, or do we prefer "provisional"?

kpet:
"Beta" is maybe more intuitive for people not familiar with our terminology but all our documents and processes use "provisional". I think I'd personally favour consistency and use "provisional". If we were to use "beta", we would need to define somewhere that "beta" means "provisional" :).

bashbaug:
I changed the name of the define to `CL_ENABLE_PROVISIONAL_EXTENSIONS`, as discussed above.

I also added some some amount of testing with the define set.  Currently, I have only added this for OpenCL 3.0 builds, to avoid doubling the number of test configurations.  My thinking is: most provisional extensions will be written against OpenCL 3.0, and OpenCL 3.0 will have all OpenCL features enabled so it'd be most likely to expose issues.  It's easy to change this, though, if other testing is desired.

EwanC (commented):
Change looks good, we should update the root `README.md` with how to opt-in

tests/CMakeLists.txt
EwanC, line 1: Could we extend the tests themselves to obverse the effects of the macro, for example something like this in`tests/test_ext_headers.c`

```cpp
int provisional_macro(void) {

#ifdef CL_ENABLE_PROVISIONAL_EXTENSIONS
   printf("%s is available\n", CL_KHR_COMMAND_BUFFER_EXTENSION_NAME);

#else

#ifdef CL_KHR_COMMAND_BUFFER_EXTENSION_NAME
   assert(false && "cl_khr_command_buffer macros should not be defined"); 
   return 1;
#endif // CL_KHR_COMMAND_BUFFER_EXTENSION_NAME
    
#endif // CL_ENABLE_PROVISIONAL_EXTENSIONS

return 0;
}


```

kpet (approved):

PR #277, opened 2025-03-17
Add definitions for cl_qcom_perf_hint

lakshmih:
Generated from cl.xml in https://github.com/KhronosGroup/OpenCL-Docs/pull/1338

### DIFF
--- a/CL/cl_ext.h
+++ b/CL/cl_ext.h
@@ -4294,23 +4294,31 @@ clCancelCommandsIMG(
 
 #endif /* !defined(CL_NO_NON_ICD_DISPATCH_EXTENSION_PROTOTYPES) */
 
-/*************************************
- * cl_qcom_perf_hint extension *
- *************************************/
+/***************************************************************
+* cl_qcom_perf_hint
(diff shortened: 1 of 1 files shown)


CL/cl_ext.h
```diff
@@ -4294,6 +4294,36 @@ clCancelCommandsIMG(
 
 #endif /* !defined(CL_NO_NON_ICD_DISPATCH_EXTENSION_PROTOTYPES) */
 
+/*************************************
+ * cl_qcom_perf_hint extension *
```
EwanC, line 4298: Is there a matching PR that adds this to the https://github.com/KhronosGroup/OpenCL-Docs/blob/main/xml/cl.xml file? Otherwise we'll need to maintain this diff whenever the headers are automatically generated from the xml.

This is also missing things that would automatically get generated from the xml, like the macros for the extension name and extension version.

bashbaug (dismissed):

EwanC (dismissed):

bashbaug (approved):
Merging as discussed over Slack.

PR #280, opened 2025-05-27
Add unions for ICD2 tags in cl_icd_dispatch

karolherbst:
This requires C11/C++11 as support for anonymous unions got added there.

I need this, because otherwise rustc doesn't let me use a constant in place of a function pointer. But I think being more explicit here would make the code a bit cleaner for loader implementors.


karolherbst:
rust bug for reference: https://github.com/rust-lang/rust/issues/141660

karolherbst:
mhhh.. not a big fan of replicating the same code block, but maybe this can be moved into a macro or some sorts. Will play around with it later.

Kerilk (approved):

bashbaug (approved):
LGTM

I'm a little nervous about losing some compiler-specific warning suppression, but if the macros are working well for the other anonymous structs, they should work fine here, too.

bashbaug:
Merging as discussed in the June 17th teleconference.

PR #289, opened 2025-08-26
Regenerate cl_ext.h from OpenCL-Docs XML

EwanC:
Use `gen_headers.py` script to regenerate the headers based on changes from OpenCL-Docs that updated cl.xml

* https://github.com/KhronosGroup/OpenCL-Docs/pull/1411
* https://github.com/KhronosGroup/OpenCL-Docs/pull/1435
* https://github.com/KhronosGroup/OpenCL-Docs/pull/1408

### DIFF
--- a/CL/cl_ext.h
+++ b/CL/cl_ext.h
@@ -54,7 +54,7 @@ extern "C" {
     "cl_khr_command_buffer"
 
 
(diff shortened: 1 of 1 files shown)


bashbaug:
I wonder if it'd be possible to add automation that automatically creates a PR with the re-generated headers when the XML file is updated...

PR #290, opened 2025-09-03
regenerate extension headers with support for cl_khr_unified_svm

bashbaug:
Regenerates extension headers after merging https://github.com/KhronosGroup/OpenCL-Docs/pull/1452.

Also includes the layer change from https://github.com/KhronosGroup/OpenCL-Docs/pull/1444.

Compared to the cl_khr_unified_svm branch, the only difference is the extension version, which is now v0.9.0 (vs. v0.2.0).


EwanC (approved):

bashbaug:
Merging as discussed in the September 9th teleconference.

PR #293, opened 2025-10-02
Add macros for anonymous union support

joselopeqti:
Introduce `__CL_HAS_ANON_UNION__` and `__CL_ANON_UNION__` for anonymous union support.
Anonymous unions are supported in C++11 and later versions, but not anonymous structs.
This change explicitly enables anonymous unions for C++11 or later, otherwise, it will fall back to the previous anonymous structs logic.

### DIFF
--- a/CL/cl.h
+++ b/CL/cl.h
@@ -139,11 +139,11 @@ typedef struct _cl_image_desc {
     size_t                  image_slice_pitch;
     cl_uint                 num_mip_levels;
     cl_uint                 num_samples;
-#if defined(CL_VERSION_2_0) && __CL_HAS_ANON_STRUCT__
-    __CL_ANON_STRUCT__ union {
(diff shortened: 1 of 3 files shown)


Kerilk:
Thanks for the contribution, I think this is a good idea.
It would require to update the loader code as well:
https://github.com/KhronosGroup/OpenCL-ICD-Loader/blob/02134b05bdff750217bf0c4c11a9b13b63957b04/loader/icd.h#L185-L189
```C
#if __CL_HAS_ANON_STRUCT__
#define ICD_ANON_UNION_INIT_MEMBER(a) {a}
#else
#define ICD_ANON_UNION_INIT_MEMBER(a) a
#endif
```

karolherbst:
> @karolherbst Would this renaming change cause a problem with the way you use the headers for rusticl?

No, the only thing that matters is how the structs look like after preprocessing and this change doesn't change that as the bindings are generated in C mode, not C++.

PR #294, opened 2025-10-07
Add function types and function pointer types for callbacks.

Kerilk:
This would be a solution to provide https://github.com/KhronosGroup/OpenCL-Docs/pull/947 without update to the XML.

### DIFF
--- a/CL/cl_ext.h
+++ b/CL/cl_ext.h
@@ -788,6 +788,11 @@ clGetMutableCommandInfoKHR(
 #define CL_APPLE_SETMEMOBJECTDESTRUCTOR_EXTENSION_VERSION CL_MAKE_VERSION(0, 0, 0)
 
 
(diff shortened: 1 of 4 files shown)


oddhack:
I saw this go by on the CL agenda and will note that in Vulkan XML, we declare function pointer types using the same XML syntax as command tags, then refer to them like any other type, e.g.

```
<type category="funcpointer">
    <proto><type>void</type> <name>PFN_vkInternalAllocationNotification</name></proto>
    <param><type>void</type>* <name>pUserData</name></param>
</type>
<type category="struct" name="VkAllocationCallbacks">
    <member><type>void</type>* <name>pUserData</name></member>
    <member><type>PFN_vkInternalAllocationNotification</type> <name>pfnInternalAllocation</name></member>
</type>
```

This makes the type information as accessible as any other struct / command parameter type and re-uses the existing header generation code (but CL might need to do something else, as I don't think you're using the Vulkan script infrastructure for some of your targets).

PR #299, opened 2026-01-21
generate headers with cl_intel_kernel_allocations_info

bashbaug:
Header support for cl_intel_kernel_allocations_info, see: https://github.com/KhronosGroup/OpenCL-Docs/pull/1522

I'll keep this as a draft until the spec/XML changes are merged.

### DIFF
--- a/CL/cl_ext.h
+++ b/CL/cl_ext.h
@@ -4672,6 +4672,28 @@ typedef cl_bitfield         cl_context_safety_properties_img;
 #endif /* defined(CL_ENABLE_BETA_EXTENSIONS) */
 
 /***************************************************************
+* cl_intel_kernel_allocations_info
+***************************************************************/
+#define cl_intel_kernel_allocations_info 1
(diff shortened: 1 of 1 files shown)


bashbaug:
@ivvenevt mind giving this a quick review?

ivvenevt:
LGTM! 👍

Kerilk (approved):
LGTM, as long as this was intentionally left out of
https://github.com/KhronosGroup/OpenCL-Headers/blob/dfb5284280a7a0bce6a21c057184ef653d9e21c1/scripts/cl_ext.h.mako#L51

bashbaug:
> as long as this was intentionally left out of [...]

Thanks, yeah, this was intentional.  In general, I think we can stop adding to the list, and just let extensions be generated in alphabetical order.

Now that we have been using the header generation toolchain for a while, IMHO it'd be nice to see of we could remove some (if not all) extensions from the list and order them alphabetically instead.  This would make it much easier to find an extension in the headers, though the downside is a pretty big (one-time) diff.

PR #302, opened 2026-02-13
Remove __stdcall warnings in ARM

joselopeqti:
In ARM architectures the `__stdcall` calling convention is ignored.

### DIFF
--- a/CL/cl_platform.h
+++ b/CL/cl_platform.h
@@ -23,31 +23,23 @@
 extern "C" {
 #endif
 
-#if defined(_WIN32)
-    #if !defined(CL_API_ENTRY)
-        #define CL_API_ENTRY
-    #endif
-    #if !defined(CL_API_CALL)
-        #if !defined(__aarch64__) && !defined(__arm64__)
-            #define CL_API_CALL __stdcall
-        #else
-            #define CL_API_CALL
-        #endif
-    #endif
-    #if !defined(CL_CALLBACK)
-        #define CL_CALLBACK     __stdcall
-    #endif
+#if defined(_WIN32) && !defined(__aarch64__) && !defined(__arm64__)
+    #define _CL_STD_CALL __stdcall
 #else
-    #if !defined(CL_API_ENTRY)
-        #define CL_API_ENTRY
-    #endif
-    #if !defined(CL_API_CALL)
-        #define CL_API_CALL
-    #endif
-    #if !defined(CL_CALLBACK)
-        #define CL_CALLBACK
-    #endif
+    #define _CL_STD_CALL
 #endif
+
+#if !defined(CL_API_ENTRY)
+    #define CL_API_ENTRY
+#endif
+#if !defined(CL_API_CALL)
+    #define CL_API_CALL     _CL_STD_CALL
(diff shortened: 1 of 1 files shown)


joselopeqti:
@bashbaug @Kerilk

CL/cl_platform.h
```diff
@@ -28,7 +28,11 @@ extern "C" {
...
     #if !defined(CL_API_CALL)
-        #define CL_API_CALL     __stdcall
+        #if !defined(__aarch64__) && !defined(__arm64__)
+            #define CL_API_CALL __stdcall
+        #else
+            #define CL_API_CALL
+        #endif
     #endif
     #if !defined(CL_CALLBACK)
         #define CL_CALLBACK     __stdcall
```
Kerilk: Shouldn't this be applied here as well, and then maybe the logic should be at a higher level.

CL/cl_platform.h
```diff
@@ -23,26 +23,20 @@
...
+#endif
+
+#if !defined(CL_API_ENTRY)
+    #define CL_API_ENTRY
+#endif
+#if !defined(CL_API_CALL)
+    #define CL_API_CALL _CL_STD_CALL
+#endif
+#if !defined(CL_CALLBACK)
+    #define CL_CALLBACK _CL_STD_CALL
```
Kerilk: I don't like the naming of the intermediate macro, since it may mislead the reader to think `stdcall`is used everywhere. Here is a suggestion:

edit: the code suggestion looks strange, don't know if it will apply.

```suggestion
#if defined(_WIN32) && !defined(__aarch64__) && !defined(__arm64__)
    #define _CL_CALL_CONV __stdcall
#else
    #define _CL_CALL_CONV
#endif

#if !defined(CL_API_ENTRY)
    #define CL_API_ENTRY
#endif
#if !defined(CL_API_CALL)
    #define CL_API_CALL _CL_CALL_CONV
#endif
#if !defined(CL_CALLBACK)
    #define CL_CALLBACK _CL_CALL_CONV
```

bashbaug: Identifiers that begin with an underscore and a capital letter are reserved, also, so we need to pick a different name than `_CL_whatever`.

https://en.cppreference.com/w/c/language/identifier.html

I really wonder if we're over-complicating this, though.  Why don't we just use:

```c
#if !defined(CL_API_CALL)
    #if defined(_WIN32) && !defined(__aarch64__) && !defined(__arm64__)
        #define CL_API_CALL __stdcall
    #else
        #define CL_API_CALL
    #endif
#endif
```

We could either do the same for `CL_CALLBACK`, or `#define` it to `CL_API_CALL` if it is not already defined.

joselopeqti: Agree

bashbaug (approved):
LGTM.

Out of curiosity, which compiler toolchain uses `__arm64__`?

joselopeqti:
> Out of curiosity, which compiler toolchain uses `__arm64__`?

AFAIA Apple Darwin

bashbaug:
Merging as discussed in the March 17th teleconference.

PR #305, opened 2026-05-14
Add cl_khr_cooperative_matrix

svenvh:
Generated from https://github.com/KhronosGroup/OpenCL-Docs/pull/1533


bashbaug:
Discussed in the May 26th teleconference.  We're going to hold off merging this for a bit longer, either until we get a few more implementers, or until the benefits of having support in the upstream headers become clearer.

I'll move this PR to a draft for now, so it doesn't get merged unintentionally.

